Issue #28, opened 2025-02-15
Packege Conflicts in Environment Installation


YangLH49:
When I use conda for the _pip install -e_ . , package conflicting occur
_hydra-core 1.3.2 requires antlr4-python3-runtime==4.9., but you have antlr4-python3-runtime 4.7.2 which is incompatible._
_omegaconf 2.3.0 requires antlr4-python3-runtime==4.9., but you have antlr4-python3-runtime 4.7.2 which is incompatible._

When I use conda for the  _pip install antlr4-python3-runtime==4.9.*_  , it said
_latex2sympy2 1.9.1 requires antlr4-python3-runtime==4.7.2, but you have antlr4-python3-runtime 4.9.3 which is incompatible._

The configuration of my device is PyTorch 2.5.1, Python 3.12 (on Ubuntu 22.04), and Cuda 12.4.

lambda7xx:
incompatible may be not a problem. can you run the script?

YangLH49:
I can't ,it said _ couldn't connect to 'https://huggingface.co' to load this file_，maybe this is a network problem rather than the environment setup problem.

michaelzhiluo:
Its just warning logs... the code still works even with the warning ;) 

If you have a fix to resolve this dependency conflict, would be really grateful 🙏

Issue #17, opened 2025-02-12
Environment Installation Problems


hushyangqing:
I have encountered a lot problem during environment installation.

1. whether conda or python virtual env should be used
2. When I use conda for the pip install -e ./verl, **disk quota exceeded** error occur. So I turn to pip install --no-cache-dir -e ./verl. However for the command pip install -e ., **package conflicting** occur

hydra-core 1.3.2 requires antlr4-python3-runtime==4.9.*, but you have antlr4-python3-runtime 4.7.2 which is incompatible.
omegaconf 2.3.0 requires antlr4-python3-runtime==4.9.*, but you have antlr4-python3-runtime 4.7.2 which is incompatible.
But there are other packages that need antlr4-python3-runtime 4.7.2.

3. When I used python virtual env, I encountered other version problem

ERROR: Could not find a version that satisfies the requirement flash_attn==2.7.3 (from deepscaler) (from versions: 0.2.0, 0.2.1, 0.2.2, 0.2.3, 0.2.4, 0.2.5, 0.2.6.post1, 0.2.7, 0.2.8, 1.0.0, 1.0.1, 1.0.2, 1.0.3, 1.0.3.post0, 1.0.4, 1.0.5, 1.0.6, 1.0.7, 1.0.8, 1.0.9, 2.0.0.post1, 2.0.1, 2.0.2, 2.0.3, 2.0.4, 2.0.5, 2.0.6, 2.0.6.post2, 2.0.7, 2.0.8, 2.0.9, 2.1.0, 2.1.1, 2.1.2.post3, 2.2.0, 2.2.1, 2.2.2, 2.2.3.post2, 2.2.4, 2.2.4.post1, 2.2.5, 2.3.0, 2.3.1.post1, 2.3.2, 2.3.3, 2.3.4, 2.3.5, 2.3.6, 2.4.0.post1, 2.4.1, 2.4.2, 2.4.3.post1, 2.5.0, 2.5.1.post1, 2.5.2, 2.5.3, 2.5.4, 2.5.5, 2.5.6, 2.5.7, 2.5.8, 2.5.9.post1, 2.6.0.post1, 2.6.1, 2.6.2, 2.6.3, 2.7.0.post2, 2.7.1.post4, 2.7.2.post1, 2.7.3, 2.7.4.post1)
ERROR: No matching distribution found for flash_attn==2.7.3

Could you please provide more detailed installation steps?

michaelzhiluo:
[...]

PR #96, opened 2025-04-17
Build with uv

aokellermann:
This PR changes the build system to use `uv` which enables reproducible builds. There are a couple issues open such as #17 and #28 related to installation/dependency issues, which should be solved by using `uv`. Now, after cloning the repo, you should only need to run

```bash
uv sync --extra build
uv sync --extra build --extra compile
```

and everything should work properly.


michaelzhiluo:
Lmk when this PR is ready for review happy to merge this!

PR #118, opened 2025-05-12
Fix Sandbox Error: TypeError - cannot pickle '_thread.lock' object

WindChimeRan:
How to reproduce:
```
python rllm/tools/code_tools/local_tools.py
```
Output:
```
Synchronous result:
CodeToolOutput(name='local_python', output=None, error="Sandbox Error: TypeError - cannot pickle '_thread.lock' object", stdout=None, stderr=None)

Asynchronous result:
CodeToolOutput(name='local_python', output=None, error="Sandbox Error: TypeError - cannot pickle '_thread.lock' object", stdout=None, stderr=None)
```
Problem:

- ProcessPoolExecutor failed with "cannot pickle '_thread.lock' object" error
- This occurred because pool.submit() attempted to pickle the entire class instance
- The instance contained the ProcessPoolExecutor itself, creating an unpicklable circular reference

Solution:

- Convert _execute_in_subprocess to static method
- Pass tool name as explicit parameter instead of via self
- Call static method via class name (PythonInterpreter._execute_in_subprocess)
- Eliminates need to pickle instance when dispatching to worker processes

### DIFF
--- a/rllm/tools/code_tools/local_tool.py
+++ b/rllm/tools/code_tools/local_tool.py
@@ -63,7 +63,8 @@ class PythonInterpreter(CodeTool):
         """
         try:
             # Submit the job to the process pool and wait for its result
-            future = self.pool.submit(self._execute_in_subprocess, code, timeout)
+            future = self.pool.submit(PythonInterpreter._execute_in_subprocess, code, timeout, self.name)
+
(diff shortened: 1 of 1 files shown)


michaelzhiluo:
Looks good, thanks!

PR #135, opened 2025-07-08
Make sure agent/env arguments are forwarded

sidnarayanan:
Hi, thank you for open sourcing a very clean code base! I was trying to run the `math_tools` example following the [README](https://github.com/agentica-project/rllm/tree/main/examples/math_tool#training), when I noticed that the tool schemas were not being passed to either the agent or the environment (see screenshot below). It appears that `init_envs_and_agents` ignores the `{env,agent}_args` passed to the init.

This is fixed by this PR, but perhaps I am missing the bigger picture. 

image


michaelzhiluo:
Awesome, thanks! I think you're right

PR #142, opened 2025-07-14
fix <tool_calls_begin> variable

wj-Mcat:
The variable of tool_calls_begin variable seems wrong, so fix it.


michaelzhiluo:
Thanks @wj-Mcat !

PR #144, opened 2025-07-16
Fix not registered license from code

annyan09023:
There are mismatches of License in the code (MIT) and License provided by the repo (Apache 2.0), remove the wrong license in the code.


michaelzhiluo:
Done ;)

PR #179, opened 2025-08-07
Add rllm_env/ to .gitignore to ignore local virtual environment

yayashuxue:


yayashuxue:
think this is an mistake generated by ai - please ignore

PR #185, opened 2025-08-13
[RLLM] Support mutli-agent PPO training

ArnavBalyan:
### Summary
Adds a lightweight multi-agent orchestration layer that executes workflows (DAG of agents).
Workflows are DAGs of agents that can be defined with some agents (nodes), connections (edges), and phases (execution schedule).
The MultiAgentExecutionEngine runs the DAG per env step, passes upstream outputs as chain context, and returns tokenized trajectories suitable for PPO. CoE (Chain of experts) is supported, MoE (Mixture of experts) is coming soon.
 
### Key concepts:
- DAG Core:
  - `AgentRole` (Proposer, Specialist, Judge, etc.)
  - `AgentConfig` (class, args, sampling params)
  - `WorkflowConnection(from_agent, to_agent, transform_fn=None)`
  - `WorkflowPhase(phase_id, agent_ids, execution_mode="sequential", ...)`
  - `BaseWorkflow` that can support `ChainOfExpertsWorkflow` (linear A→B→C) etc.

- **Engine**
  - `MultiAgentExecutionEngine` orchestrates phases, injects chain context, drives env with the final agent’s action, and emits PPO-ready batches.
  - Context injection per edge with optional `transform_fn`.

- **Trainer**
  - `MultiAgentPPOTrainer` on top of existing PPO.
  - `training_mode`: 
    - `final_agent` (default): train only last agent’s tokens/masks.
    - `unified`: concat all agents’ prompts/responses (heavier).

Factory: Can be used to create other workflows in the future, MoE or 3D graph of agents or other usecases. 

**Note**: Currently we support final_agent, i.e final agent's output can be used for policy update. This can be much more interesting in the future, such as rewarding individual agents in the same chain with weighted rewards (depending on each agent's performance / contribution).

This involves multiple agents each managing their own chain; significantly heavy on resources and can oom on 32 batch size on 2x A100s. We should consider maybe fast offloading kv cache and retrieval in the future

image

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,4 @@
+[submodule "verl"]
+	path = verl
+	url = https://github.com/agentica-project/verl
+	branch = main


ArnavBalyan:
cc @michaelzhiluo @jeffreysijuntan

PR #186, opened 2025-08-13
[RLLM] Switch verl remote to SSH

ArnavBalyan:
- Moving verl remote from http to ssh for security/convenience

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,4 @@
 [submodule "verl"]
 	path = verl
-	url = https://github.com/agentica-project/verl
+	url = git@github.com:agentica-project/verl.git
 	branch = main


ArnavBalyan:
cc @michaelzhiluo @jeffreysijuntan minor change could you PTAL thanks

jeffreysijuntan:
Since verl is public, what's the main purpose of doing this? If you switch from https to verl, then user needs to configure an ssh key with Github beforehand to get it cloned, is that right?

ArnavBalyan:
Was to offer some convenience if making submodule changes, might be easier for developers if they have keys configured. We can probably keep as-is if accessibility for one time cloners is important

PR #209, opened 2025-09-11
Add VimGolf agent training example

James4Ever0:
From #204

### DIFF
--- a/examples/vimgolf/lib.py
+++ b/examples/vimgolf/lib.py
@@ -25,9 +25,7 @@ class VimGolfSingleTurnAgent(BaseAgent):
         self.messages = []
         self.accumulate_thinking = accumulate_thinking
 
-    def update_from_env(
-        self, observation: Any, reward: float, done: bool, info: dict, **kwargs
-    ):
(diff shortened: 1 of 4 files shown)


jeffreysijuntan:
Thank you for your contribution! Can you run `pre-commit run --all-files` to apply ruff formatter and then commit again with the changes? 

Also, it would be nice to add this as an example to our official doc, so ppl can reference it. If you have existing training run results would be nice to share them there as well.

jeffreysijuntan:
Still seems to fail the check.. What command did you run locally?

James4Ever0:
> Still seems to fail the check.. What command did you run locally?

In the previous commit, I have tried to install precommit dependencies with uv, but stuck at flash-attn.

I then ran the following command according to [.pre-commit-config.yaml](https://github.com/James4Ever0/rllm/blob/main/.pre-commit-config.yaml):

```bash
cd examples/vimgolf
ruff check --fix --show-fixes --output-format=full
```

I guess there was something missing. I will run the following command afterwards.

```bash
cd examples/vimgolf
isort *.py
black *.py
ruff check --fix --show-fixes --output-format=full
ruff format
```
Is there a way to install only pre-commit dependencies (ruff) with uv?

PR #228, opened 2025-09-29
Standalone inference: remove hard verl dependency

JasonWei05:
- Make verl optional using lazy imports and TYPE_CHECKING.
- Add vllm and sglang to pyproject.toml.
- Inference runs without verl, and training is unchanged when verl is installed.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -74,6 +74,10 @@ dependencies = [
     "PyYAML",
     "pydantic",
 
+    # Inference backends
+    "vllm",
+    "sglang",
+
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -74,6 +74,10 @@ dependencies = [
     "PyYAML",
     "pydantic",
 
+    # Inference backends
+    "vllm",
```
jeffreysijuntan: Let users install these themselves. We just expose a OpenAI endpoint and don't care about the underlying serving engine.

PR #230, opened 2025-09-30
[feat] support multimodal

abrohamLee:
Still debugging. It seems the model didn't see the multimodal info


rllm/engine/rollout/verl_engine.py
yuzeng0-0, line 1: At line 191, where `output_batch = self.rollout_manager.generate_sequences(batch)` is called, an asynchronous runtime error occurs.
I was wondering if you have encountered this issue before.

Issue #196, opened 2025-08-24
about lora support


zjcanjux:
it seems add following script in bash 
```python
     ...
    actor_rollout_ref.model.lora_rank=64 \
    actor_rollout_ref.model.lora_alpha=32  \

    ...

```

will support lora ?

jeffreysijuntan:
We use verl as our backend, so if the current verl supports LoRA, then yes. You can refer to our `v0.2` branch where we integrate with the latest verl v0.5, and you should be able to use features officially supported by verl.

Issue #141, opened 2025-07-12
LoRA support for vLLM V1


Danau5tin:
Hello 👋

First of all, I want to sat a huge thank you for this repo! 
I have tried a couple more which have large bugs, however rLLM is stable which is amazing, and the Agent/Env APIs are exactly what i was hoping for.

I can run my training run just fine (tried a single epoch, all went well).
However I saw from this epoch that my training would be expensive 😅, so I opted to go for LoRA.

When I switched to LoRA, the training crashes on the first step, and after some investigation, I believe it might be because there is new syntax for LoRA in vLLM V1, and as I explain below, I cannot use vLLM v0. 

## Error logs
```
(train_agent pid=40730)  Trajectory Reward=0.03
[...]

PR #239, opened 2025-10-09
[Draft] Fix async LoRA support for vLLM v1 + FSDP

listar2000:
## Description

This PR aims to patch several `verl` related issues that prevent LoRA training with vLLM V1 + FSDP (as requested in #196, #141). This has been a lingering issue (e.g. mentioned [here](https://github.com/volcengine/verl/issues/3271) and [here](https://github.com/volcengine/verl/issues/2048)) for `verl` and affect the version (0.5.0) that `rLLM v0.2` is pinning. Specifically, there are **two crashing issues** and **one hidden bug**. The two crashing issues are:

1. **[`verl/workers/sharding_manager/fsdp_vllm.py`](https://github.com/volcengine/verl/blob/8fdc4d3f202f41461f4de9f42a637228e342668b/verl/workers/sharding_manager/fsdp_vllm.py#L294)**  
   In `FSDPVLLMShardingManager.update_params`, the call to `self.inference_engine.llm_engine.add_lora(lora_request)` fails because, in async mode, `self.inference_engine` no longer has a `llm_engine` attribute. This has been explicitly pointed out [here](https://github.com/volcengine/verl/issues/2048).

2. **[`verl/workers/rollout/vllm_rollout/vllm_rollout_spmd.py`](https://github.com/volcengine/verl/blob/8fdc4d3f202f41461f4de9f42a637228e342668b/verl/workers/rollout/vllm_rollout/vllm_rollout_spmd.py#L408)**  
   In `vLLMAsyncRollout`, the `lora_kwargs` field within `kwargs` isn’t propagated, causing the inference engine to initialize without a `lora_manager` during `init_worker`.  

_Note: these issues have been addressed in the newest `main` branch of the verl respository. Our patches simply apply those newer fixes to our pinned verl version._

During the patching of the above issues, I've also found out that our previous way of monkey patching (#237) through **post-hoc reassignment** (e.g. `Cls.old_method = new_method`) is not as maintainable and robust as **leveraging the `wrap_function_wrapper` in `wrapt`**. So some refactorings for the `rllm/patches/verl_patch_hook.py` are also done.
  
## Changes Made
- `Verl` patches for the two critical errors above.
- Additional special handling of `vLLM`'s `max_lora_rank` configuration, which originally prevented us from specifying `lora_rank < 8` in our training script (the low-rank setting is gaining much attention lately, see [this blogpost](https://thinkingmachines.ai/blog/lora/))
- Refactoring of the patches module.
- A new script `train_hendrycks_math_lora.sh` under `examples/simple_math` with `lora_rank = 1` (testing right now).

## TODOs
- [ ] Pass the pytests.
- [x] Use pre-commit to format.
- [ ] Implement some checks & warnings so users can be more informed (e.g. the actor and rollout model dtypes must match, and the maximum lora rank cannot exceed 512, etc.)
- [x] Tested locally (with `examples/frozenlake/train_frozenlake_agent.sh` and `lora_rank = 8`) with multi-GPU training setup. See [wandb](https://wandb.ai/listar2000/rllm-agent) for some initial steps.
- [x] Test the very low lora rank (`lora_rank = 1`) setting (see below for the verifications done).


rllm/patches/patch_vllm_async_server.py
```diff
@@ -0,0 +1,185 @@
...
+    prompt_ids = ba["prompt_ids"]
+    sampling_params = ba["sampling_params"]
+    request_id = ba["request_id"]
+
+    max_tokens = instance.max_model_len - len(prompt_ids)
+    sampling_params = SamplingParams(max_tokens=max_tokens, **sampling_params)
+    prompt = TokensPrompt(prompt_token_ids=prompt_ids)
+
+    lora_request = LoRARequest(lora_name=f"{LORA_ID_MAGIC}", lora_int_id=LORA_ID_MAGIC, lora_path=LORA_PATH_MAGIC)
+    generator = instance.engine.generate(prompt=prompt, sampling_params=sampling_params, request_id=request_id, lora_request=lora_request)
```
listar2000, line 170: This is the crucial place where (before the patch) the generate function still makes request to the base model (not the LoRA model) in the AsyncLLM engine. Essentially this makes the RL training off-policy as the rollout is always done by the (not optimized) base model when LoRA is on.

listar2000:
**Update:** this PR will be delayed to a future version of `rllm`. The patching of `verl` seems unnecessarily complicated so I opened a PR for `verl` https://github.com/volcengine/verl/pull/3821 instead.

PR #256, opened 2025-10-20
make rllm-specific configs applied correctly and robustly

listar2000:
This PR addresses several (potential) issues related to applying rLLM-specific configurations — `disable_thinking` and `accumulate_reasoning`:

- Fixes #255 by adding the missing key (`accumulate_reasoning`) to `agent_ppo_trainer_megatron.yaml`.
- Improves robustness by allowing `accumulate_reasoning` to be absent in custom config files, defaulting to `False` when missing.
- Corrects `AsyncAgentExecutionEngine` initialization, which previously attempted to read `disable_thinking` from `kwargs`. This field is reserved for `config.rllm.agent.engine_args` (often empty). The fix ensures `disable_thinking` is now properly read from the `rllm` config with appropriate fallback handling.

_Note: Extra care was taken to ensure compatibility with cases where users reuse `verl` configs that may omit the `config.rllm` fields (while other fields are usually present in these cases)_

PR #258, opened 2025-10-21
Ensure disable_thinking defaults to False when config is None

Tendo33:
refer : https://github.com/rllm-org/rllm/pull/256#issuecomment-3425376077


listar2000:
Thanks for the catch! This should resolve the issues that #256 introduced to all the examples' `run_xxx.py` scripts.

Issue #259, opened 2025-10-21
Bug Report: safe_import fails due to circular import between env_agent_mappings and single_turn_workflow


yubin1991:
**Describe the bug**

When initializing workflow class mappings in env_agent_mappings.py, the import of SingleTurnWorkflow fails and returns None.
As a result, WORKFLOW_CLASS_MAPPING only contains one entry instead of all expected workflows.

To Reproduce

Run the environment setup that includes env_agent_mappings.py

Observe the console output:

in env agent mapping 1 {'single_turn_workflow': None}

The safe_import() function silently fails to load SingleTurnWorkflow.

Expected behavior
All workflow classes (SingleTurnWorkflow, MultiTurnWorkflow, SimpleWorkflow, CumulativeWorkflow) should be successfully imported and included in WORKFLOW_CLASS_MAPPING.

Root cause analysis
This issue is caused by a circular import:

rllm.workflows.single_turn_workflow
    → imports rllm.trainer.env_agent_mappings
        → calls safe_import("rllm.workflows.single_turn_workflow")

During the second import, single_turn_workflow is only partially initialized, so the SingleTurnWorkflow class is not yet defined, resulting in an AttributeError.

Proposed fix
Move the import of AGENT_CLASS_MAPPING and ENV_CLASS_MAPPING inside the __init__ method of SingleTurnWorkflow to delay the import until runtime.
Example:

class SingleTurnWorkflow(Workflow):
    def __init__(self, agent_cls, env_cls, agent_args=None, env_args=None, **kwargs):
        super().__init__(**kwargs)
        from rllm.trainer.env_agent_mappings import AGENT_CLASS_MAPPING, ENV_CLASS_MAPPING
        agent_cls = AGENT_CLASS_MAPPING[agent_cls] if isinstance(agent_cls, str) else agent_cls
        env_cls = ENV_CLASS_MAPPING[env_cls] if isinstance(env_cls, str) else env_cls
        self.agent = agent_cls(**(agent_args or {}))
        self.env = env_cls(**(env_args or {}))

PR #261, opened 2025-10-22
fix: circular import issues in WORKFLOW_CLASS_MAPPING

listar2000:
This PR addressses the issue mentioned in #259.

The fix is straightforward: for any workflow that imports from `env_agent_mappings.py`, (i.e. `single_turn_workflow`, `multi_turn_workflow`, and `cumulative_workflow`), we delay the lazy import inside the workflow `__init__` function.

Note: even though the original issue only mentions the missing of `single_turn_workflow`, the issue actually causes the **first workflow without lazy importing to throw an error and return `None`** -- so we need to patch for all workflows involved.


jeffreysijuntan:
Thanks for the fix!

Issue #262, opened 2025-10-22
ValueError: Using V1 AsyncLLMEngine, but envs.VLLM_USE_V1=False.


LianShuQuan:
# Command: bash rllm/examples/swe/train_deepswe_32b.sh
I followd the set up of 
```
# Clone the repository
git clone --recurse-submodules https://github.com/rllm-org/rllm.git
cd rllm

# Create a conda environment
conda create -n rllm python=3.10 -y
conda activate rllm

# Install verl
bash scripts/install_verl.sh

# Install rLLM
pip install -e .

pip install -e ./R2E-Gym

pip install swebench==3.0.2

```

## But got error while already export VLLM_USE_V1=1.

```
  File "xxx/envs/rllm/lib/python3.10/concurrent/futures/_base.py", line 458, in result
    return self.__get_result()
  File "xxx/rllm/lib/python3.10/concurrent/futures/_base.py", line 403, in __get_result
    raise self._exception
  File "xxx/rllm/lib/python3.10/site-packages/verl/workers/rollout/vllm_rollout/vllm_async_server.py", line 267, in init_engine
    self.engine = AsyncLLM.from_vllm_config(vllm_config)
  File "xxx/rllm/lib/python3.10/site-packages/vllm/v1/engine/async_llm.py", line 143, in from_vllm_config
    raise ValueError(
ValueError: Using V1 AsyncLLMEngine, but envs.VLLM_USE_V1=False. This should not happen. As a workaround, try using AsyncLLMEngine.from_vllm_config(...) or explicitly set VLLM_USE_V1=0 or 1 and report this issue on Github.

Set the environment variable HYDRA_FULL_ERROR=1 for a complete stack trace.
```

## some dependency:
```
vllm                      0.8.5.post1              pypi_0    pypi
torch                     2.6.0                    pypi_0    pypi
transformers              4.51.1                   pypi_0    pypi
flash-attn                2.7.4.post1              pypi_0    pypi

xformers                  0.0.29.post2             pypi_0    pypi
xgrammar                  0.1.18                   pypi_0    pypi


opentelemetry-api         1.38.0                   pypi_0    pypi
opentelemetry-exporter-otlp 1.26.0                   pypi_0    pypi
opentelemetry-exporter-otlp-proto-common 1.26.0                   pypi_0    pypi
opentelemetry-exporter-otlp-proto-grpc 1.26.0                   pypi_0    pypi
[...]

PR #265, opened 2025-10-24
Fix environment variable forwarding to ray runtime env

listar2000:
#### What is this PR about?

Fix issue #262 

Specifically, this PR now correctly **overrides** (instead of **dropping**) the default `PPO_RAY_RUNTIME_ENV` in `ray_runtime_env.py` when we export certain environment variables from script. We add a default `VLLM_USE_V1 = 1` as we are now pinning `verl = 0.5.0` (which uses a version of `vllm` that embraces the V1 engine).

With the suggestion of @kylemontgomery1, for a more complete fix, we will also forward any driver environment variable with the following prefixes are forwarded:

- **Inference Engines**: `VLLM_`, `SGL_`, `SGLANG_`
- **HuggingFace Libraries**: `HF_`, `TOKENIZERS_`, `DATASETS_`
- **Training Frameworks**: `TORCH_`, `PYTORCH_`, `DEEPSPEED_`, `MEGATRON_`
- **CUDA/NCCL**: `NCCL_`, `CUDA_`, `CUBLAS_`, `CUDNN_`, `NV_`, `NVIDIA_`

We further let the user to specify a flag `RLLM_EXCLUDE` to rule out any prefix or particular variable that **the user wants to exclude from the above forwarding**. As an example:

```bash
export RLLM_EXCLUDE="VLLM*,CUDA*,NCCL_IB_DISABLE"
# Excludes all VLLM_*, all CUDA_*, and the specific NCCL_IB_DISABLE variable
```
Corresponding tests and the documentation of the `ray_runtime_env` module is also added.

---
This PR also makes sure that `ray_init_settings` can be properly passed into `train_workflow_pipeline` (so that it's now consistent with `train_agent_ppo`.

### DIFF
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -33,6 +33,7 @@ The engine module contains the core execution infrastructure:
 The trainer module provides RL training capabilities:
 
 - **Agent Trainer**: Main training interface for RL algorithms
+- **Ray Runtime Environment**: Configuration for Ray runtime environment
 
 ### 🛠️ Tools
 The tools module provides a comprehensive framework for creating and managing tools:
--- a/docs/api/trainer/ray_runtime_env.md
+++ b/docs/api/trainer/ray_runtime_env.md
@@ -1,0 +1,111 @@
+# Ray Runtime Environment Configuration
+
+## Overview
+
+The `ray_runtime_env` module automatically forwards relevant environment variables from your local environment to Ray worker processes during distributed training. This ensures that configuration for libraries like VLLM, NCCL, CUDA, and HuggingFace are properly propagated to all workers.
+
+## Environment Variable Forwarding
+
+### Automatic Forwarding
+
+Environment variables with the following prefixes are automatically forwarded to Ray workers:
+
+- **Inference Engines**: `VLLM_`, `SGL_`, `SGLANG_`
+- **HuggingFace Libraries**: `HF_`, `TOKENIZERS_`, `DATASETS_`
+- **Training Frameworks**: `TORCH_`, `PYTORCH_`, `DEEPSPEED_`, `MEGATRON_`
+- **CUDA/NCCL**: `NCCL_`, `CUDA_`, `CUBLAS_`, `CUDNN_`, `NV_`, `NVIDIA_`
+
+### Default Environment Variables
+
+The following variables are set by default for PPO training:
+
+```python
+{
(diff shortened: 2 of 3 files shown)


kylemontgomery1:
@listar2000 For a complete fix, maybe we can forward the relevant variables from the driver process (in order to ensure env variables not already in PPO_RAY_RUNTIME_ENV get forwarded).

```python
PPO_RAY_RUNTIME_ENV = {
    "env_vars": {
        "TOKENIZERS_PARALLELISM": "true",
        "NCCL_DEBUG": "WARN",
        "VLLM_LOGGING_LEVEL": "WARN",
        "VLLM_ALLOW_RUNTIME_LORA_UPDATING": "true",
        "CUDA_DEVICE_MAX_CONNECTIONS": "1",
        "VLLM_USE_V1": "1",
    },
    "worker_process_setup_hook": "rllm.patches.verl_patch_hook.setup",
}

FORWARD_PREFIXES = (
    "VLLM_", "SGL_", "SGLANG_", 
    "HF_", "TOKENIZERS_", "DATASETS_",
    "TORCH_", "PYTORCH_", "DEEPSPEED_", "MEGATRON_", 
    "NCCL_", "CUDA_", "CUBLAS_", "CUDNN_", "NV_", "NVIDIA_",
)

def get_ppo_ray_runtime_env():
    env = PPO_RAY_RUNTIME_ENV["env_vars"].copy()
    forwarded = {
        k: v for k, v in os.environ.items()
        if any(k.startswith(p) for p in FORWARD_PREFIXES)
    }
    env.update(forwarded)
    return {
        "env_vars": env,
        "worker_process_setup_hook": PPO_RAY_RUNTIME_ENV["worker_process_setup_hook"],
    }
```

listar2000:
The idea LGTM, but I wonder whether it's possible that this will introduce unwanted side effects (e.g. the user is setting some driver variables not intended for ray, or even for rLLM)?

listar2000:
@kylemontgomery1 
Bear with me 😂 for making this PR heavier. My recent commit integrates your suggestion, while also adding a `RLLM_EXCLUDE` flag to let the user decide what env var (either excluding with prefix matching with `VLLM*`, or excluding a particular variable name directly).

Since the logic is more complicated (while IMO more robust now), I've added some tests and the documentation for this module as well.

kylemontgomery1:
Looks good to me. Thanks!

PR #268, opened 2025-10-25
[nightly] update recent changes on workflow engines

listar2000:
#### What is this PR about?

This PR merges several changes from my local `nightly` branch that makes sure the `agent_workflow_engine` is compatible with `Verl`'s 0.6.x new APIs. A good sanity check is the successful running of the following experiments on **LoRA x solver-judge-workflow** [wandb log](https://wandb.ai/listar2000/solver-judge-workflow).

Specifically:

- `RayPPOTrainer` now removes the `_validate_config` method (leaving the validation to the `__post_init` method in the many new config classes (e.g. [`ActorConfig`](https://github.com/volcengine/verl/blob/main/verl/workers/config/actor.py)) 
-  I've modified the init argument of `AgentWorkflowEngine` to reflect the current status quo that this class is now **highly coupled** with `VerlEngine` -- e.g. other engines like `FireworksEngine` do not have `wake_up` or `sleep` methods at all. We might want to have some major refactoring later as we start to embrace non-`Verl` backends.


rllm/engine/agent_workflow_engine.py
```diff
@@ -22,7 +23,7 @@
 
 
 class AgentWorkflowEngine:
-    def __init__(self, workflow_cls: type[Workflow], workflow_args: dict, rollout_engine: RolloutEngine, config=None, n_parallel_tasks: int = 128, retry_limit: int = 3, raise_on_error: bool = True, **kwargs):
+    def __init__(self, workflow_cls: type[Workflow], workflow_args: dict, rollout_engine: VerlEngine, config=None, n_parallel_tasks: int = 128, retry_limit: int = 3, raise_on_error: bool = True, **kwargs):
```
jeffreysijuntan: Are there any incompatibilities with the RolloutEngine? We want to keep the general RolloutEngine class to support inference using different engines.

listar2000: I will roll back this line. I guess the issue now is that following the API change, the best way to do `wake_up`/`sleep` for the `VerlEngine` is through an async function, e.g.
```python
async def wake_up(self):
    """Wake up all rollout replica instances asynchronously."""
    await asyncio.gather(*[replica.wake_up() for replica in self.rollout_manager.rollout_replicas])
```
But our original definition for `RolloutEngine` was still defining a sync version of `wake_up`/`sleep` (I guess we should probably make these methods async anyway since `execute_tasks_verl` is already async).

jeffreysijuntan: Great. Yes I think they can be async as well. Right now I think function is only used for VerlEngine only, so  whatever you did here will not affect other classes.

PR #271, opened 2025-10-26
fix controlling the n_parallel_agents and the concurrent env operations

LianShuQuan:
Does `n_parallel_agents` means the number of active agents, and `max_workers` means the concurrent env operations?
If so, then
`trajectory_generator()` should use `asyncio.Semaphore()` like `execute_tasks()`;
`self.executor = ThreadPoolExecutor(max_workers=max_concurrency)` should be modified in `trajectory_generator()`;
`self.n_parallel_agents = len(envs)` should be deleted in `update_envs_and_agents()`.

### DIFF
--- a/rllm/engine/agent_execution_engine.py
+++ b/rllm/engine/agent_execution_engine.py
@@ -30,7 +30,7 @@ class AgentExecutionEngine:
         tokenizer=None,
         rollout_engine=None,
         chat_parser=None,
-        n_parallel_agents=1,
+        n_parallel_agents=1,  # The number of active agents
         trajectory_timeout=None,
         gamma=0.2,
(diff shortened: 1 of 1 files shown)


jeffreysijuntan:
Great catch. How about renaming `max_workers` to `max_env_workers` then?

LianShuQuan:
> Great catch. How about renaming `max_workers` to `max_env_workers` then?

Sure. Should I only change `max_workers` to `max_env_workers` besides the initialization parameters in order to maintain compatibility with the initialization and usage code of AgentExecutionEngine?

jeffreysijuntan:
Yes, simply rename this parameters to `max_env_workers` to make it clearer.

PR #275, opened 2025-10-27
Fix broken prompt length filtering

erranlli:
### Problem
- Dataset registration used placeholder prompts in VERL format
- VERL's `filter_overlong_prompts` filtered placeholders instead of actual task questions
- Long prompts (>max_prompt_length) caused training crashes at runtime

### Solution
- Modified `apply_verl_postprocessing()` to use actual task questions
- Added better error logging for prompt length issues
- Updated comments to explain when length errors can occur

### Breaking Change
**Important**: If resuming from old checkpoints, delete `data.pt` from checkpoint directories:
```bash
find checkpoints/ -name "data.pt" -delete
```
Old checkpoints contain cached datasets with placeholder prompts.

### DIFF
--- a/rllm/data/dataset.py
+++ b/rllm/data/dataset.py
@@ -385,8 +385,13 @@ class DatasetRegistry:
         """
         processed_data = []
         for entry in data:
+            # Extract the actual question for VERL filtering to work correctly
+            # Try common field names: question, prompt, instruction, input
+            question = entry.get("question") or entry.get("prompt") or entry.get("instruction") or entry.get("input") or "placeholder"
+
             processed_entry = {
-                "prompt": [{"role": "user", "content": "placeholder"}],
+                # Use actual question so VERL's length filtering works correctly
+                "prompt": [{"role": "user", "content": question}],
                 "reward_model": {
                     "style": "rule",
                     "ground_truth": None,
--- a/rllm/engine/agent_execution_engine.py
+++ b/rllm/engine/agent_execution_engine.py
@@ -207,8 +207,12 @@ class AgentExecutionEngine:
         messages = agent.chat_completions
         prompt_tokens, _ = convert_messages_to_tokens_and_masks(messages, tokenizer=self.tokenizer, parser=self.chat_parser, contains_first_msg=True, contains_generation_msg=True)
         prompt_token_len = len(prompt_tokens)
-        # Note, this should never happen!
+        # This can happen if:
+        # 1. Initial task prompt from environment is too long (should be filtered during dataset loading)
+        # 2. Checkpoint contains cached dataset that wasn't filtered (delete checkpoint's data.pt)
         if prompt_token_len > self.max_prompt_length:
+            # Log details for debugging
+            logger.warning(f"Trajectory {idx}: Initial prompt length {prompt_token_len} exceeds max_prompt_length {self.max_prompt_length}. First 200 chars of prompt: {self.chat_parser.parse(messages[:1], add_generation_prompt=False)[:200]}...")
             agent.reset()
             raise Exception(f"Trajectory {idx}: initial prompt length {prompt_token_len} already exceeded max_prompt_length {self.max_prompt_length}, retrying")
 


rllm/data/dataset.py
```diff
@@ -385,8 +385,13 @@ def apply_verl_postprocessing(cls, data: list[dict[str, Any]]) -> list[dict[str,
         """
         processed_data = []
         for entry in data:
```
jeffreysijuntan, line 387: We don't need to make any modifications here. "prompt" is never used in rLLM during training, we make a placeholder just to be compatible with `verl`'s data structure.

erranli: @jeffreysijuntan That is correct. However, Verl needs it for dataset filtering when data.filter_overlong_prompts=true The placeholder makes filtering logic incorrect.  Below is the call flow.
train_deepscaler_8k.sh
  │
  └─> python3 -m examples.deepscaler.train_deepscaler
       │
       ├─> main() decorated with @hydra.main
       │   └─> AgentTrainer.train()                 # Line 26 in train_deepscaler.py
       │       │
       │       └─> ray.get(TaskRunner.run.remote())  # Line 80 in agent_trainer.py
       │           │
       │           └─> TaskRunner.run()               # Line 62 in train_agent_ppo.py
       │               │
       │               ├─> AgentPPOTrainer.__init__()  # Line 195 in train_agent_ppo.py
       │               │
       │               ├─> trainer.init_workers()      # Line 209
       │               │
       │               └─> trainer.fit_agent()         # Line 211
       │                   │
       │                   └─> for batch_dict in self.train_dataloader:  # Line 156 in agent_ppo_trainer.py
       │                       │
       │                       │ (First iteration triggers lazy initialization)
       │                       │
       │                       └─> RayPPOTrainer._init_dataloader()  # Called by parent class
       │                           │
       │                           └─> create_rl_dataset()  # Line 566 in ray_trainer.py
       │                               │  (from verl.trainer.main_ppo import create_rl_dataset)
       │                               │
       │                               └─> RLHFDataset.__init__()  # Line 284 in main_ppo.py
       │                                   │  dataset = dataset_cls(data_files=..., tokenizer=..., config=...)
       │                                   │
       │                                   ├─> self._download()                    # Line 120 in rl_dataset.py
       │                                   │
       │                                   └─> self._read_files_and_tokenize()     # Line 121
       │                                       │
       │                                       └─> self.maybe_filter_out_long_prompts()  # Line 140
       │                                           │
       │                                           └─> dataframe.filter(lambda doc: doc2len(doc) <= max_prompt_length)  # Lines 169-173

Key Files in Order
1. examples/deepscaler/train_deepscaler_8k.sh → Launches training
2. examples/deepscaler/train_deepscaler.py → Creates AgentTrainer and calls .train()
3. rllm/trainer/agent_trainer.py → Launches Ray TaskRunner
4. rllm/trainer/verl/train_agent_ppo.py → TaskRunner.run() creates AgentPPOTrainer
5. rllm/trainer/verl/agent_ppo_trainer.py → fit_agent() iterates over self.train_dataloader
6. verl/verl/trainer/ppo/ray_trainer.py → _init_dataloader() calls create_rl_dataset()
7. verl/verl/trainer/main_ppo.py → create_rl_dataset() instantiates RLHFDataset
8. verl/verl/utils/dataset/rl_dataset.py → RLHFDataset.__init__() calls maybe_filter_out_long_prompts()

The filtering happens before the first training batch is yielded, during dataset initialization when AgentPPOTrainer.fit_agent() first iterates over self.train_dataloader.

at rl_dataset.py:
def doc2len(doc) -> int:
    return len(tokenizer.apply_chat_template(doc[prompt_key], add_generation_prompt=True))

dataframe = dataframe.filter(
    lambda doc: doc2len(doc) <= self.max_prompt_length,
    num_proc=self.num_workers,
    desc=f"Filtering prompts longer than {self.max_prompt_length} tokens",
)

jeffreysijuntan: We don't want to use this feature from verl because we also have no idea of what becomes the "prompt" in the user's data structure, as this is dependent on the user. Setting `question = entry.get("question") or entry.get("prompt") or entry.get("instruction") or entry.get("input") or "placeholder"` does not fundamentally solve this problem. 

Instead, it is better on the rLLM side to handle it. Right now we just throw an Exception, but a better approach is to not throw an Exception and skip it if it becomes too long. That way it does not break training.

PR #280, opened 2025-10-31
fix: Gracefully skip overlong prompts during training to prevent crashes

erranlli:
## Summary
Implements graceful degradation for prompts exceeding `max_prompt_length`, preventing training crashes.

## Problem
Training crashed when prompts exceeded max length: Exception: Trajectory {idx}: initial prompt length 3302 already exceeded max_prompt_length 2048, retrying

## Solution
- ✅ Overlong prompts return `None` and are skipped gracefully
- ✅ Batch size dynamically adjusts to match

## Key Changes
1. **`agent_execution_engine.py`**: Return `None` for overlong prompts instead of crashing
2. **`agent_ppo_trainer.py`**: Track skipped indices and filter batch to match

## Benefits
- Training continues instead of failing
- No NaN gradients from division by zero
- Dynamic batch size adjustment
- Clean and simple implementation

## Testing
```bash
bash examples/deepscaler/test_graceful_degradation.sh
```

Expected: Training continues with warnings when 3302-token prompt is encountered, no crashes.

## Files Changed
- `rllm/engine/agent_execution_engine.py` (graceful degradation)
- `rllm/trainer/verl/agent_ppo_trainer.py` (batch alignment)

### DIFF
--- a/rllm/engine/agent_execution_engine.py
+++ b/rllm/engine/agent_execution_engine.py
@@ -437,6 +437,10 @@ class AgentExecutionEngine:
             self.executor = ThreadPoolExecutor(max_workers=self.max_env_workers)
         semaphore = asyncio.Semaphore(self.n_parallel_agents)
 
+        # Initialize skipped indices and valid indices lists (will be populated as trajectories complete)
+        self._last_skipped_indices = []
+        self._last_valid_indices = None  # Will be computed after all trajectories complete
+
         if self.engine_name == "verl":
             self.rollout_engine.wake_up()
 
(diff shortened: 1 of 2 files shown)


LianShuQuan:
in generate_agent_trajectories_async() 
```
                async for item in self.agent_execution_engine.trajectory_generator(timing_raw=timing_raw, mode=mode, meta_info=meta_info):
                    # This item can not be None. Instead, overlong prompts will be skipped.
                    queue.put(item)
```
Because 
```
            if item is None:
                break
```


And then in generate_agent_trajectory()
```
                for trajectory in gen_seq_generator:
                    # Skip None trajectories (overlong prompts)
                    if trajectory is not None:
                        trajectories.append(trajectory)
```
is not necessary

Issue #291, opened 2025-11-05
No actor/rollout weight sync when free_cache_engine=False


sneng1997:
Hi,
I am running rllm v0.2 on a custom dataset and found that there is no weight sync between the actor and rollout models if free_cache_engine is set to False. Is this a designed behavior or something to be fixed?

Thanks in advance

listar2000:
Hi, thanks for pointing this out. I think currently the check for `free_cache_engine` is indeed unnecessary within the `agent_execution_engine` and `agent_workflow_engine` -- and having `free_cache_engine = False` would possibly cause the issues you've mentioned (have you experienced this in experiments? I'm running some experiments as well).

I would say this seems like something that `Verl` also fails to address, but I would experiment with a tentative fix soon and issue a PR if the fix is effective.

listar2000:
Hi. As a follow-up I've drafted a PR with a tentative fix. However this PR is targeting the `nightly` branch of `rLLM`. If you are using `v0.2`, feel free to take a look at the changes I've made there and patch it manually.

Overall I would say this issue seems to be ignored within `verl` as well (but maybe they are using `sync` rollout mostly so this is not a problem). It is generally recommended to turn `free_cache_engine = True`.

sneng1997:
Thanks for the prompt fix!

PR #298, opened 2025-11-09
[Draft] Fix free cache engine weight sync

listar2000:
This is a tentative fix to the issue #291: where the weight sync between rollout engine and trainer engine **is completely missing** (an obvious issue that `verl` also hasn't fixed yet) when config `actor_rollout_ref.rollout.free_cache_engine=False`.

Fortunately this seems fixable without patching `verl` as in `rLLM` we implement our own rollout trajectory generation logic. In fact, the `rollout_engine.wake_up()/sleep()` method should be run **regardless of the `free_cache_engine`** flag -- the flag will be checked again anyway during the `wake_up/sleep` call anyway. For instance, the release of `kv-cache` in the inference engine will do an extra check [here](https://github.com/volcengine/verl/blob/60d8a62c0d810be1c3e9134c14ac188c20342a42/verl/workers/fsdp_workers.py#L698).

This PR should **not affect the usual case when `free_cache_engine = True` at all**. Some experiments are on-going to see if there is any side-effect when `free_cache_engine = False` (so this PR is still under draft)

### DIFF
--- a/rllm/engine/agent_execution_engine.py
+++ b/rllm/engine/agent_execution_engine.py
@@ -13,6 +13,7 @@ from rllm.agents.utils import (
     convert_messages_to_tokens_and_masks,
     get_recent_assistant_user_messages,
 )
+from rllm.engine.rollout.rollout_engine import ModelOutput
 from rllm.environments.base.base_env import BaseEnv
 from rllm.environments.env_utils import (
     compute_mc_return,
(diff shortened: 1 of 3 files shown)


listar2000:
Before these changes get merged, I've filed a similar PR in `verl`: 
https://github.com/volcengine/verl/issues/4147

In case they have a better solution.

listar2000:
This has been merged in a newer version of this change in #324.

PR #301, opened 2025-11-10
Fireworks API returns raw token ids to avoid re-tokenization

1stprinciple:
Replace OpenAIClient with curl

### DIFF
--- a/rllm/trainer/config/agent_ppo_trainer.yaml
+++ b/rllm/trainer/config/agent_ppo_trainer.yaml
@@ -86,7 +86,7 @@ fireworks:
   deployment_id: null
   model_id_prefix: test-model
   concurrency: 32
-  return_token_ids: False
+  return_token_ids: True
 
 trainer:
   log_episodes: false


rllm/trainer/config/agent_ppo_trainer.yaml
```diff
@@ -65,7 +65,8 @@ fireworks:
   deployment_id: null
   model_id_prefix: test-model
   concurrency: 32
+  return_token_ids: False
```
jeffreysijuntan: Let's just set it directly to True? And then up there we don't need to specify it in `train_fireworks_math.sh` explicitly any more.

examples/fireworks_math/train_fireworks_math.sh
```diff
@@ -77,5 +78,6 @@ python3 -m examples.fireworks_math.train_fireworks_math \
     trainer.default_hdfs_dir=null \
     trainer.total_epochs=100 \
     rllm.workflow.use_workflow=True \
-    fireworks.deployment_id=wtk15cs9 \
-    fireworks.model_id_prefix=qwen3-4b-math
\ No newline at end of file
+    fireworks.deployment_id=rllm-qwen3-8b-3 \
+    fireworks.model_id_prefix=rllm-qwen3-8b-math \
+    fireworks.return_token_ids=True
```
jeffreysijuntan, line 83: We can set `return_token_ids` to True by default so no need to specify it here.

rllm/trainer/config/agent_ppo_trainer.yaml
1stprinciple: Done.

PR #308, opened 2025-11-16
[nightly] merge recent updates in main back to nightly

listar2000:
## What is this PR about?

This PR tries to merge some recent updates/bugfixes in the `main` branch to the ongoing `nightly` branch. The purpose is to prevent these two branches from diverting too much (except for new features under implementation in `nightly`). A few notable additions include:

- Bugfix that only 1 GPU is running #307 
- Episode logger #282
- Removing `old_log_probs` recalculation #304 

among many others.

---
This PR also resolves a few merge conflicts -- most of them are straight-forward and obviously compatible. The only one noticeable merge is on the [agent_trainer.py](https://github.com/listar2000/rllm/blob/2a09c8ddb3bc446da824221c88a0fe5ff61b33de/rllm/trainer/agent_trainer.py), where the `main` version supports `["verl", "fireworks"]` backends, while the current `nightly` supports `["verl", "tinker"]` -- and the resolution is to support all three of them.


jeffreysijuntan:
nice, thanks for making the merge.

listar2000:
> nice, thanks for making the merge.

No problem. Anything else needed before the merge?

jeffreysijuntan:
We found a bug in one of the commit that affects dataset loading, and have it reverted  [here](https://github.com/rllm-org/rllm/commit/574fa4a41c634669638bf92b00bb974b7c854cb5). It seems like this is not reverted here?

listar2000:
@jeffreysijuntan Thanks for the catch. It turns out that this is due to **I was merging with the `upstream/main` as a remote-tracking branch and forgot to `fetch` newer commits first**. Just remerge with most up-to-date commits on `main`.

Issue #346, opened 2025-12-14
install_verl.sh uses flash-attn==3.12 which does not support Python 3.11 (incompatible with conda env setup instructions


weiyali126:
### 🐛 Bug Report

The installation instructions recommend creating a Python 3.11 conda environment:

```bash
conda create -n rllm python=3.11 -y
```

However, the script install_verl.sh installs:
```bash
pip install --no-cache-dir flash_attn-2.8.0.post2+cu12torch2.7cxx11abiFalse-cp312-cp312-linux_x86_64.whl
```

❗ Problem

Are the "python version" and "torch version" written incorrectly?

listar2000:
@kylemontgomery1 Hi Kyle, is the `install_verl.sh` script currently deprecated (I see it was last updated ~ 2 months ago)?

kylemontgomery1:
Yeah, install_verl.sh should be removed from main since it's specific to the v0.2.0 release, and specifically verl==0.5.0. @listar2000 Do you mind opening a PR? 

For installing rllm==v0.2.1 with verl==0.6.1, run:
```
uv venv --python 3.11
uv pip install -e .[verl] 
```

To address the original question, the python/torch version mismatch shouldn't introduce any issues, but you can always grab the correct wheel from [here](https://github.com/Dao-AILab/flash-attention/releases/tag/v2.8.0.post2) or build flash-attn yourself i.e., `pip install flast-attn --no-build-isolation`.

listar2000:
The new README should give a clearer installation guide.

PR #347, opened 2025-12-15
[Doc] Update README and fix a few installation related issues

listar2000:
### What does this PR do?

- It addresses the issue in #346 by providing a cleaner README installation guide (despite we already have a detailed version in our docs). Also fix a few places like wrong `v0.2.1` release time, etc.
- It removes the legacy `install_verl.sh` to avoid confusion.
- It also adds the legacy `verl/` folder to `.gitignore` (in case users upgrade from an older version and might encounter this folder not being able to get removed).

### DIFF
--- a/rllm/engine/agent_sdk_engine.py
+++ b/rllm/engine/agent_sdk_engine.py
@@ -16,13 +16,13 @@ from tqdm import tqdm
 from rllm.agents.agent import Episode, Trajectory
 from rllm.engine.rollout import ModelOutput, RolloutEngine
 from rllm.engine.rollout.verl_engine import VerlEngine
-from rllm.utils import colorful_print
 from rllm.sdk.data_process import group_steps, trace_to_step
 from rllm.sdk.protocol import Trace, TrajectoryView
 from rllm.sdk.proxy.proxy_manager import VerlProxyManager
 from rllm.sdk.session import SESSION_BACKEND
 from rllm.sdk.session.base import wrap_with_session_context
 from rllm.sdk.store.sqlite_store import SqliteTraceStore
+from rllm.utils import colorful_print
 from rllm.workflows.workflow import TerminationReason
 
 # Avoid hard dependency on verl at import time; only for typing


README.md
```diff
@@ -37,52 +37,51 @@ rLLM is an open-source framework for post-training language agents via reinforce
 
 ## Getting Started 🎯
 
-### Step 1: Building rLLM
+rLLM requires `Python >= 3.11`. You can install it either directly via pip or build from source.
```
kylemontgomery1: 3.10 is fine, only tinker requires 3.11.

README.md
```diff
@@ -37,52 +37,51 @@ rLLM is an open-source framework for post-training language agents via reinforce
...
 
-rLLM requires Python >= 3.11. You can install it either directly via pip or build from source.
+There are three ways that you can install rLLM:
 
-**Option A: Direct Installation**
+### Approach A: Direct Installation
 
 ```bash
-uv pip install "git+https://github.com/rllm-org/rllm.git"
+uv pip install "rllm[verl] @ git+https://github.com/rllm-org/rllm.git"
```
kylemontgomery1, line 47: not sure if there is any use case for a direct installation @jeffreysijuntan 
and if we do we should change DatasetRegistry to cache the datasets and register file to some user location like ~./rllm

README.md
```diff
@@ -37,52 +37,51 @@ rLLM is an open-source framework for post-training language agents via reinforce
...
+# Create an uv environment
+uv venv --python 3.11
+source .venv/bin/activate
 ```
 
-### Step 2: Installing Training Backend
+**Step 2: Install rLLM with Training Backend**
 
-rLLM supports two training backends: `verl` and `Tinker`. Choose one based on your needs.
+rLLM supports two training backends: `verl` and `tinker`. Choose one based on your needs.
```
kylemontgomery1, line 68: they can always install both e.g., uv pip install -e .[verl,tinker]

README.md
```diff
@@ -37,52 +37,51 @@ rLLM is an open-source framework for post-training language agents via reinforce
...
-**Option A: Using `verl` as Training Backend**
+_**Option I:** Using `verl` as Training Backend_
 
 ```bash
-# Initialize and update the verl submodule
-git submodule update --init --recursive
-
-# Install verl
-bash scripts/install_verl.sh
+# can add --torch-backend=auto to automatically detect the CUDA backend
```
kylemontgomery1: in my testing, 'auto' did not work, ideally the user would always set this e.g., even in the cpu case they can set --torch-backend=cpu

README.md
listar2000: Yeah but I assume in most cases ppl would stick to only one of them maybe?

listar2000:
Thx for the review and great suggestions @kylemontgomery1. Since this PR relates to the crucial README, I would wait and see if @jeffreysijuntan has any comments (& response to some of the points you've raised). Will add a followup commit tmr.

README.md
jeffreysijuntan: I think we can just enforce the user to use 3.11 now, given that Tinker has this requirement. This make sure user don't have to switch when they want to switch engine.

README.md
jeffreysijuntan: Currently no, but I think it is a good idea to cache datasets to ~/.rllm anyway so we should support this feature.

README.md
jeffreysijuntan: let's remove this comment then.

PR #348, opened 2025-12-15
Unblock sdk installation by overriding dependices

wht0703:
Resolved `uv` resolution errors caused by conflicting transitive dependencies.

Fixed: `eval-protocol` was forcing an outdated `litellm version (1.9.4)` that lacked proxy support.

Fixed: `smolagents` and `litellm` had conflicting requirements for the `rich` library.

Changes applied via `[tool.uv] override-dependencies` to force compatible versions of `litellm` and `rich`. Verified with `uv sync --all-extras`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -121,6 +121,12 @@ code-tools = [
     "together>=1.4",
 ]
 
+[tool.uv]
+override-dependencies = [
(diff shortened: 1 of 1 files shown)


listar2000:
LGTM

PR #361, opened 2026-01-08
Merge unified trainer interface into a new nightly branch

listar2000:
## What is this PR about?

On a high-level, this PR will pin down a few high-level features in rLLM v0.3, including

1. A [`UnifiedTrainer`](https://github.com/rllm-org/rllm/blob/unified-trainer/rllm/experimental/unified_trainer.py) that controls the agentic RL training loop main logic, and each `backend` will implement the [`BackendProtocol`](https://github.com/rllm-org/rllm/blob/unified-trainer/rllm/experimental/protocol.py) interface.
2. A [`UnifiedWorkflowEngine`](https://github.com/rllm-org/rllm/blob/unified-trainer/rllm/experimental/engine/unified_workflow_engine.py) that becomes **backend-agnostic** -- i.e. implementing a standard 
  `execute_tasks(self, tasks: list[dict], task_ids: list[str]...) -> list[Episode]`
  function to generate rollouts of episodes.
3. A new advantage and policy loss calculation module ([`common/advantage.py`](https://github.com/rllm-org/rllm/blob/unified-trainer/rllm/experimental/common/advantage.py)) that attempts to standardize advantage estimator and RL algorithms across different `backends`.
4. A new configuration system that separates `rLLM`-specific parameters (e.g. configs for the training loop like `num_epochs`) away from backend-specific ones (e.g. the `actor_rollout_ref.xxx` for `Verl`). 

## Design docs
- [Discussions on episode -> trajectory group logic](https://www.notion.so/Shared-Episode-Transform-Protocol-2aa08daca1dd804396a3d74055be3b4b?source=copy_link)
- [Documentation on the `UnifiedTrainer`](https://docs.google.com/document/d/1Et3VSvZ7XdZv4y4e-Zprv-zp9e7N4_CXSuGuSppj1GU/edit?usp=sharing_)

## Incoming follow-up PRs
- @jeewoo-lee will implement the `SkyRL backend`.
- Someone will help implement more advantage estimators.
- @listar2000 will perform more testings and fixes (see below for known issues).
- @listar2000 will support training agent-env (i.e. `agent_ppo_trainer`) in the same trainer.

## Migration plan (TBD with separate PRs) into `main`

- [ ] Merge new utility function (e.g. better trajectory visualizer/logger and timer) into main.
- [ ] Merge the Tinker-related updates (including `TrajectoryGroup` data structure).
- [ ] Merge the Verl-related updates.
- [ ] Merge the `rLLM`-native advantage estimator and policy loss modules.
[...]

PR #377, opened 2026-01-23
Merge nightly branch into main

listar2000:
### What this PR is about?

Please refer to PR #361 for an **overview of the new `UnifiedTrainer` design**. Below we only **highlight all non-`experimental/` files that this PR touches** and why they **do not break existing codes**. Note that in general we try to keep a copy within `experimental/` if we have to make potentially breaking changes to any main component (e.g. the `experimental/rollout`) -- but in rare cases below we directly make changes to the original files with backward compatibility in mind.

### Changes to non-experimental components
- `rllm/agents/agent.py`: 
    - `Step`: some basic typing improvement, adding a `__post_init__` to unify the behavior of initializing the `Step` object (sometimes user just passes in `ModelOutput` -- we then fill out other fields for the user). Also adding an `advantage` field for holding the result for advantage computation -- also useful for future implementation of advantage merging.
    - `Trajectory`: change the default to `.reward` from 0 to `None` since it's possible that the user does not want to specify trajectory-level reward at all (so `None` is semantically more explicit).
    - `TrajectoryGroup`: we copy a version of this from the `tinker_data_processor.py` to here (the former will be deprecated later, but both definitions co-exist without conflicts for now).
- `rllm/engine/agent_workflow_engine.py`: 
    -  Improve the logging display logic so for each rollout: if the trajectory does not have a reward, we will display the reward for its last step instead.

### DIFF
--- a/rllm/agents/agent.py
+++ b/rllm/agents/agent.py
@@ -47,8 +47,8 @@ class Step:
 
         # check that the token ids are filled
         # TODO(listar2000): this might cause compatibility issue. Double check if we should make these assertions.
-        assert len(self.prompt_ids) > 0, "prompt_ids is empty"
-        assert len(self.response_ids) > 0, "response_ids is empty"
+        # assert len(self.prompt_ids) > 0, "prompt_ids is empty"
+        # assert len(self.response_ids) > 0, "response_ids is empty"
 
         # check that the lengths would match up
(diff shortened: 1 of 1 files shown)


listar2000:
@jeffreysijuntan Should be ready to go once you take a look at the few files that are not in the `experimental` folder.

PR #393, opened 2026-02-11
FinQA Release

mananroongta:


listar2000:
Wow congrats @mananroongta for the release. But just curious shouldn't this example go into the `recipe/`?

Issue #392, opened 2026-02-09
reward about search example


Fulin-Gao:
We have trained the search task in the examples, but there is no sign of reward improvement within 200 steps. The official example uses epoch=100 by default, which takes a huge amount of time. Could you please advise approximately after how many steps we can expect to see performance improvement or convergence?

thwu1:
Hi @Fulin-Gao , would you mind providing more details in what example you run (the path to the files)? This is definitely not normal.

Fulin-Gao:
@thwu1 Thanks for your reply. I'm running the official example at rllm/examples/search. I've tried changing the instruct model and adjusting hyperparameters, but still no upward trend in reward. Do you have any suggestions?

thwu1:
@Fulin-Gao Would you mind sharing the wandb curve?

PR #403, opened 2026-02-26
[feature] Support rLLM SDK in UnifiedTrainer

jeffreysijuntan:
We are moving forward with the unified trainer in rLLM v0.3. The goal of this PR is to integrate rLLM SDK with the unified trainer as well, so from now on we have only one trainer to maintain.

Features supported in this PR:
1. Integration rLLM SDK with OpenAI Agent SDK, Google ADK, and Strands from Amazon. For each of these agent frameworks, we add a hook to automatically transform the traces into `TrajectoryView` object in rLLM SDK. 
2. Integrate rLLM SDK's abstraction with `UnifiedTrainer` by creating a specialized adapter called `SDKWorkflow` that transforms the trajectories collected via SDK into standard `Episode` format in rLLM for further processing.
3. Add a simple `countdown` example using the Unified trainer to verify training works.

ToDO:
- []  Deprecate the original SDK trainer and engine.
- []  Refactor some existing data schema and abstractions of the SDK.
- []  Performance optimization. 
- []  Build a more comprehensive tracing solution and integrate with rllm-ui.


listar2000:
Hi @jeffreysijuntan thanks for the efforts integrating SDK to `unified_trainer`! I will start review and hopefully get the review done by the end of today (or early tmr). But a few points for now:

1. I feel like the `examples` folder is keep expanding with every commit. From a user's perspective I think having thousands lines of example scripts (unrelated to them) pouring every time doing `git pull` can be daunting. Also many examples might be outdated or having some bugs -- e.g. #392 (fixed now). Should we consider leaving only the most "useful" examples (e.g. those illustrating basic use cases and can be referenced by doc) and moving others to a separate repo?
2. I can focus on the changes you've made to the `experimental` folder, and it might be great to have @thwu1 or someone else familiar with the SDK to review SDK-related codes.

Also, before this PR gets merged there are two PRs I need to "queue earlier":
1. The @Chanbinski 's PR on #402.
2. Another feature that we discussed earlier on allowing "trajectory (role)-specific advantage computation". Will send the draft PR on that soon.

jeffreysijuntan:
Thanks for the suggestion. I do think we need a clean up of the `examples` folder now as it becomes pretty bloated

PR #405, opened 2026-02-27
Fixed the qwen parser to handle case when model is non-thinking but user put disable_thinking=false as config

jeewoo-lee:
**Problem**

When using a non-instruct (non-thinking) model with `disable_thinking=False`, the parser incorrectly treated the entire output as reasoning content. Since these models never emit `<think>` tags, the completion text was being placed into the `reasoning` field with an empty `content`, causing answer parsing to fail.

**Solution**

Updated the `parse_completion` method to distinguish between two cases when `</think>` is absent and thinking is enabled:

1. **`<think>` tag present, no `</think>`** — The model started reasoning but didn't close the tag. Treat the rest as reasoning with empty content (existing behavior).
2. **No `<think>` tag at all** — The model is a non-thinking model that doesn't produce think tags. Treat the full text as content with empty reasoning (new behavior).

This ensures non-instruct models work correctly regardless of the `disable_thinking` setting.

### DIFF
--- a/rllm/parser/chat_template_parser.py
+++ b/rllm/parser/chat_template_parser.py
@@ -506,6 +506,13 @@ class QwenChatTemplateParser(ChatTemplateParser):
 
             return self.user_token + tool_outputs_str + self.eot_token
 
+    def _strip_special_tokens(self, text):
+        if text.endswith(self.eos_token):
+            text = text[: -len(self.eos_token)]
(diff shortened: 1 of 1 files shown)


listar2000:
@kylemontgomery1 thx for helping review this.

kylemontgomery1 (approved):
lgtm!

PR #433, opened 2026-03-09
fix(rewards): add sympy timeout to prevent Ray stalls

### LINKED ISSUE #95: The Ray process is stuck, with only the main worker in a running state.
When I trained the DS-1.5B on single node with 8*80G GPUs, the program did not report any errors. However, the Ray process became stuck, no new workers were allocated (all parameters used deepscaler's default configurations), and most Python processes on the server were in the S state. Has anyone encountered this issue before?

Image

rajatbeladiya:
Fixes intermittent training stalls where Ray appears "stuck" and workers stop making progress.

In issue #95, one suspected root cause is Sympy hanging (e.g. on certain expressions) during reward evaluation. When this happens in long-running Ray workers, the whole training job can effectively stall without an obvious Python exception.

Changes
- Add a configurable timeout around Sympy parse/simplify in `are_equal_under_sympy`.
  - Uses `SIGALRM` (main-thread only) to interrupt pathological cases.
  - Timeout is controlled via `RLLM_SYMPY_TIMEOUT_S` (default: 2.0s). Set to <= 0 to disable.
- Add a regression test that simulates a hanging `sympy.simplify` and asserts the call returns quickly.
- Improve test stubbing logic so optional deps are only stubbed if not importable.

Fixes #95

### DIFF
--- a/tests/rewards/test_sympy_timeout.py
+++ b/tests/rewards/test_sympy_timeout.py
@@ -1,4 +1,3 @@
-import os
 import time
 
 


listar2000:
Thanks for the catch!

PR #443, opened 2026-03-15
fix(deps): cap transformers <5 until vLLM compatibility

### LINKED ISSUE #388: There may be some compatibility conflicts between Transformers 5.0 and older vLLM.
Recently, Transformers released version 5.0.0, which may introduce some compatibility issues.
 Our pyproject.toml currently only specifies "transformers>=4.55.0", so during installation, it might automatically install version 5.0.0.
 This could cause problems because the vLLM version that verl depends on might be too old to be compatible with Transformers 5.0.

### Specific Error:

(APIServer pid=98439)   File "/rllm/.venv/lib/python3.11/site-packages/transformers/tokenization_utils_base.py", line 1326, in __getattr__
(APIServer pid=98439)     raise AttributeError(f"{self.__class__.__name__} has no attribute {key}")
(APIServer pid=98439) AttributeError: Qwen2Tokenizer has no attribute all_special_tokens_extended. Did you mean: 'num_special_tokens_to_add'?

### Reproduction Steps:
Follow the quickstart guide, install the environment, and run rllm/examples/math_tool/run_math_with_tool.py.

Tested Working Combination:
• transformers: 4.57.0
• vLLM: 0.11.0

This combination works correctly.

### Possible Solutions:
1. Temporary Fix: Add version constraints in pyproject.toml to temporarily limit Transformers to versions below 5.0.0 (e.g., transformers>=4.57.0,<5.0.0).

2. Alternative Approach: Upgrade vLLM to a version that supports Transformers 5.0.0 (though this version is automatically installed by verl, which might require coordination).

rajatbeladiya:
Issue #388: Transformers 5.0.0 can be auto-installed due to an open-ended lower bound, but older vLLM stacks used via verl can be incompatible (e.g. tokenizer attribute errors).

This PR adds an upper bound to prevent installing Transformers v5 until the dependency stack is confirmed compatible.

Fixes #388

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,6 +29,10 @@ dependencies = [
     "openai",
     "tqdm",
     "requests",
+    # Core ML/AI packages
+    "torch",
+    "torchvision",
+    "transformers>=4.57.0,<5.0.0",
     # Data
     "datasets>=2.20.0",
     "pillow",


kylemontgomery1:
Fixed by commit `b10d9b0 `.

PR #460, opened 2026-03-25
feat: add Daytona sandbox backend for code execution

### LINKED ISSUE #459: Add Daytona as a code execution sandbox backend
### Problem / Motivation

Add Daytona as a sandbox backend for `code_tools`, alongside the existing E2B and Together backends.

Daytona provides cloud sandboxes for code execution with a simple SDK. Having it as an option gives users more flexibility in choosing their sandbox provider; different pricing, regions, and feature sets.

### Proposed Solution

- New `DaytonaPythonInterpreter` class in `rllm/tools/code_tools/` implementing the `CodeTool` interface
- Wire it into `PythonInterpreter` as `backend="daytona"`
- Add `daytona-sdk` to the `[code-tools]` optional dependency group

### Alternatives Considered

_No response_

### Additional Context

_No response_

mu-hashmi:
fixes #459

## Summary
Adds daytona as a new cloud sandbox backend. This uses the `daytona` package and the `daytona.Daytona` sdk.

## What changed
- adds `rllm/tools/code_tools/daytona_tool.py`
- wires `PythonInterpreter(backend="daytona")` to the new backend
- adds `daytona` to the `[code-tools]` extra
- updates the installation docs to mention daytona

## Implementation Notes
- uses `sandbox.process.code_run(...)` instead of the code interpreter/session path, so execution stays stateless per call and matches current `CodeTool` usage in rllm
- passes through the daytona config knobs that matter here: `api_url`, `snapshot`, and `env_vars`
- retry handling: auth errors fail fast, other errors retry
- maps successful runs to `stdout`, and non-zero exits to `stderr` plus the last traceback line in `error`

## Validation
```
$ python -m rllm.tools.code_tools.daytona_tool
CodeToolOutput(name='daytona_python',
               output=None,
               error=None,
               metadata=None,
               stdout='Hello, world!\n',
               stderr=None)
CodeToolOutput(name='daytona_python',
               output=None,
               error='ZeroDivisionError: division by zero',
               metadata=None,
               stdout=None,
               stderr='Traceback (most recent call last):\n'
                      '  File "<stdin>", line 1, in <module>\n'
                      'ZeroDivisionError: division by zero\n')
```


JiwaniZakir (commented):
In `daytona_tool.py`, the retry logic in `forward` has a subtle issue: when all retries are exhausted, `_restart_sandbox` is called to create a fresh sandbox, but then an error is immediately returned without attempting to execute `code` on the new sandbox — so the restart effort is entirely wasted for the current call. The caller receives an error and must retry manually, defeating the purpose of the restart.

The `__main__` block's async test is also broken: `interpreter(code=..., use_async=True)` passes `use_async` through `**kwargs` to `forward`, which never inspects it, so the call returns a `CodeToolOutput` synchronously rather than a coroutine — `await coro` will then raise a `TypeError`. Either `forward` needs an `async def` variant or the example should be removed.

Minor: `super().__init__()` is called after `self._init_sandbox()` in `__init__`, which means sandboxes are already allocated before the parent class is fully initialized — if the base `CodeTool.__init__` raises, you'd leak sandbox resources. Calling `super().__init__()` first would be safer.

mu-hashmi:
@JiwaniZakir updated PR addressing review comments (apologies for letting the PR go stale for this long!)

mu-hashmi:
@JiwaniZakir just wanted to gently bump this, updated the PR addressing the issues you brought up. appreciate the review!

PR #461, opened 2026-03-26
Upgrade rLLM to support Verl 0.7.1

listar2000:
## Summary

- **Upgrade Verl dependency from 0.6.1 to 0.7.1**, adapting to new checkpoint_manager API, async rollout architecture, and sleep/wake lifecycle
- **Fix NCCL timeout in old trainer** (`AgentWorkflowPPOTrainer`): move `_balance_batch` before `compute_log_prob` to prevent FSDP all-gather desync caused by unbalanced token distribution across DP ranks
- **Fix missing re-pad in VerlBackend**: add `_pad_dataproto_to_world_size` before gradient updates in `update_policy` to guard against post-filtering batch size changes
- **Migrate weight sync to checkpoint_manager**: replace deprecated vLLM sleep/wake calls with `checkpoint_manager.sleep_replicas()` / `update_weights()`
- **Per-role policy loss functions** for VerlBackend (e.g., separate loss routing for solver vs judge roles)
- **Guard multi_modal_inputs access** to prevent KeyError on non-multimodal training runs
- Various import fixes, config compatibility updates, and code cleanup for Verl 0.7.1

### Verl 0.7.1 API migration
- `checkpoint_manager.update_weights()` / `sleep_replicas()` replace old manual vLLM sleep/wake
- Updated imports for renamed/moved Verl modules
- Bumped verl version in `pyproject.toml`

### NCCL desync fix (old trainer)
- `_balance_batch` was called AFTER `compute_log_prob` — moved to BEFORE, matching Verl native `fit()` and `VerlBackend` ordering
- Without balancing, uneven token distributions across DP ranks cause different micro-batch counts under `use_dynamic_bsz=true`, leading to mismatched FSDP all-gathers

### VerlBackend re-pad fix
- Added `_pad_dataproto_to_world_size` in `update_policy` before `update_critic`/`update_actor`
- `mask_truncated_samples` filtering can shrink the batch to a non-divisible size after the initial pad

## Test plan

- [ ] Run `tmp/run_verl_solver_judge_classic.sh` (old trainer path) — verify no NCCL timeout, training completes multiple batches
- [ ] Run unified trainer path — verify no regression
- [ ] Verify multimodal training still works (multi_modal_inputs guard is backwards-compatible)
- [ ] Check training metrics (loss, entropy, reward) are reasonable in first few steps


luyuzhe111:
@listar2000 looks like changes in https://github.com/rllm-org/rllm/pull/441 is somehow removed?

listar2000:
@luyuzhe111 yeah this is weird -- won't merge it for now.

PR #475, opened 2026-04-02
Add early-finalize continuation for truncated reasoning rollouts

### LINKED ISSUE #267: [new feature request] support for truncating early if running out of context
I want to train a reasoning model with a maximum of 16k generation length.
Models (esp. thinking models) often lead to >16k generation length by default. I want to be able to truncate early (say 14k tokens) and still use the trajectory by adding a statement like "</thinking> The answer is: " at the end of the partially generated trajectory, and re-sending it to the server, and this way the trajectory will most probably complete within the next 2k tokens (making it 16k in total) or get truncated again (in which case it can be handled normally as it is handled right now)

is it possible to add this support to rllm?

thank you!

taivu1998:
## Summary

This PR adds opt-in support for early-finalizing truncated long-form generations so we can reserve answer budget inside a single completion window.

Closes #267.

### What changed

- add a workflow-level early-finalize helper that:
  - reserves a configurable tail budget from `max_tokens`
  - runs an initial generation with the reduced budget
  - if that generation stops due to length, optionally appends a synthetic suffix only when reasoning was cut off inside an unfinished `<think>...</think>` block
  - continues generation from token input for the reserved tail budget
- add token-in/token-out continuation support to the rollout stack and timing layer for the Verl path
- wire the new helper into the standard workflow implementations and the FinQA workflow
- add `Step.response_mask` so synthetic suffix tokens remain in `completion_ids` for continuity while being excluded from Verl loss
- update the Verl transform to preserve the explicit step-level response mask instead of always assuming an all-ones loss mask
- add config knobs for `rllm.early_finalize`
- add focused tests for early-finalize behavior and response-mask propagation

## Design notes

- the feature is opt-in and defaults to disabled
- v1 stays tightly scoped to the workflow/Verl path that the issue targets
- synthetic answer forcing is narrow by design: for non-thinking truncated outputs we simply continue from the partial completion without injecting a prefix
- existing prompt-length guards remain unchanged; this only handles the case where a single completion runs out of response budget

## Testing

- `python -m pytest tests/engine/test_early_finalize.py tests/unified_trainer/test_verl_transform.py tests/rewards/test_math_reward.py -q`
- `ruff check rllm/engine/rollout/rollout_engine.py rllm/engine/rollout/verl_engine.py rllm/workflows/early_finalize.py rllm/workflows/timing_mixin.py rllm/workflows/single_turn_workflow.py rllm/workflows/multi_turn_workflow.py rllm/workflows/cumulative_workflow.py rllm/agents/agent.py rllm/experimental/verl/transform.py rllm/experimental/verl/__init__.py projects/finqa/train_finqa.py tests/engine/test_early_finalize.py tests/unified_trainer/test_verl_transform.py`
- `python -m py_compile rllm/engine/rollout/rollout_engine.py rllm/engine/rollout/verl_engine.py rllm/workflows/early_finalize.py rllm/workflows/timing_mixin.py rllm/workflows/single_turn_workflow.py rllm/workflows/multi_turn_workflow.py rllm/workflows/cumulative_workflow.py rllm/agents/agent.py rllm/experimental/verl/transform.py rllm/experimental/verl/__init__.py projects/finqa/train_finqa.py`

### DIFF
--- a/projects/finqa/train_finqa.py
+++ b/projects/finqa/train_finqa.py
@@ -4,6 +4,7 @@ from rllm.agents.agent import Episode
 from rllm.data.dataset import DatasetRegistry
 from rllm.engine.rollout.rollout_engine import ModelOutput
 from rllm.trainer.agent_trainer import AgentTrainer
+from rllm.workflows.early_finalize import attach_model_output_to_step, maybe_generate_with_early_finalize
 from rllm.workflows.multi_turn_workflow import MultiTurnWorkflow
 from rllm.workflows.workflow import TerminationEvent, TerminationReason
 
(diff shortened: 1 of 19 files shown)


kylemontgomery1:
I lean towards this being implemented at the workflow level (e.g., SingleTurnWorkflowWithEarlyFinalize) instead of a global feature in rLLM. @listar2000 What do you think?

listar2000:
Will take a look thx @kylemontgomery1 for pointing this to me.

taivu1998:
Hi @kylemontgomery1, @listar2000, could you help review again? Thanks!

PR #481, opened 2026-04-03
Integrate fully async training to UnifiedTrainer

kylemontgomery1:
## Summary

Adds fully async training to UnifiedTrainer, currently only for tinker backend.

## Type of change

- [x] Feature
- [ ] Fix
- [ ] Docs
- [ ] Refactor
- [ ] Example / Project
- [ ] Infra / CI

## What changed

- New SyncCoordinator, TrajectoryGroupBuffer, and MetricsAggregator for async pipeline                                                      
- Async training loop in UnifiedTrainer with concurrent generation/training and coordinator-managed weight sync                                                                                                                                     
- on_policy_updated hook added to protocal for weight sync/checkpointing                                                                      
- Countdown examples for sync and async unified tinker training   

## Validation

- [x] Not run (reason below)

Validation details:
Entire repo needs linting. Tested both paths on wandb: https://wandb.ai/agentica/rllm-countdown

None

## Docs / examples

- [ ] Not needed
- [ ] Updated docs
- [ ] Updated examples
- [x] Follow-up docs needed

## Related issues / PRs

- Fixes #
- Related to #
- Stacked on / depends on #


listar2000:
Hi @kylemontgomery1 just want to point you on a few recent PRs to main on the (non-fully-async) Verl support in rLLM: #474 #483 

These two PRs both help rLLM support the new engine worker (for `VerlBackend`, we no longer supports the more legacy workers, i.e. we now force `self.config.trainer.use_legacy_worker_impl == "disable"`.

I feel that the fully async Verl support will go with a parallel, and hopefully non-conflicting, path or backend. But just in case -- maybe check how these recent PRs might or might not disrupt your ongoing work and if so I can help take a look at where the conflicts happen.

rllm/agents/agent.py
```diff
@@ -50,6 +55,7 @@ def info(self, value: dict) -> None:
         self.metadata = value
 
     def model_post_init(self, __context: Any) -> None:
+        self.chat_completions = deepcopy(self.chat_completions)
```
kylemontgomery1, line 58: This deepcopy was incorrectly removed during the refactor from dataclasses to pydantic. Many old workflows operate with:

```python
for turn in max_turns:
    output: ModelOutput = await self.rollout_engine.get_model_response(messages)
    messages.append("role": "assistant", "content": output.content, ...}
    trajectory.steps.append(Step(chat_completions=messages, model_output=output))
```

If chat completions is not deepcopied, then appending a message on a future turn would mutate a previous turn's step.chat_completions.

listar2000: I guess in the future we should really have a rLLM built-in `messages` format & class (similar to Tinker's `Message`), and ensure (1) it's as easy to work with as a plain dictionary, while (2) every step only holds a "view" of it (so no need to keep lots of copies, while earlier steps are not affected).

rllm/experimental/rollout/rollout_engine.py
luyuzhe111, line 7: have been seeing this type_checking thing a lot in the codebase. curious if it's actually needed?

listar2000: There are mostly for static type checking purpose (code readeability, LSP support, etc.) without actually importing the files during runtime for overhead (often used when we just need the type of a variable). I would say they are in general helpful, but there can be cases where we can discard them (e.g. class type too obvious or can be directly inferred).

listar2000:
@kylemontgomery1 All those annoying pre-commit issues, failing tinker engine test, etc. have been resolved in some latest PRs to the main. #487

rllm/agents/agent.py
kylemontgomery1: Agreed, I think we can spend some time this week rethinking messages/parsers.

listar2000:
@kylemontgomery1 Merged and great efforts! I think we should plan for some blogpost and definitely doc updates once the Verl's side is also ready.

PR #488, opened 2026-04-05
refactor: move experimental rollout engines to canonical rllm/engine/rollout/

listar2000:
## Summary
- Graduate rollout engines from `rllm/experimental/rollout/` to `rllm/engine/rollout/` (RolloutEngine, VerlEngine, TinkerEngine, types, completer)
- Update 24+ consumer files to import from the canonical location
- Add backward-compat shim in `rllm/experimental/rollout/__init__.py` with `DeprecationWarning`
- Rename `.validate` → `.is_validation` across rollout engines (4 files, 8 occurrences)

## Test plan
- [x] `pytest tests/engine/` passes
- [x] `pytest tests/parser/` passes
- [x] Importing from `rllm.experimental.rollout` still works (with deprecation warning)
- [x] End-to-end GPU test with Tinker backend
- [ ] End-to-end GPU test with Verl backend


tests/engine/test_tinker_engine.py
```diff
@@ -1,203 +1,202 @@
...
-        id=id,
-    )
+
+def _make_image_chunk(length: int = 16):
+    """Create a real tinker ImageChunk with the given token length."""
+    return ImageChunk(data=b"\x89PNG", format="png", expected_tokens=length)
 
 
 # ------------------------------------------------------------------
-# _convert_openai_messages
```
luyuzhe111: is this a merge error? why are all tool use related tests removed?

tests/engine/test_tinker_engine.py
```diff
@@ -1,203 +1,202 @@
...
-    _parse_tinker_message,
-    _prepare_messages_with_tools,
+from rllm.engine.rollout.tinker_engine import (
+    _flat_token_input_length,
+    _flat_token_input_to_model_input,
 )
-from rllm.tools.tool_base import ToolCall as RllmToolCall
 
 # ------------------------------------------------------------------
-# Fixtures
```
luyuzhe111, line 18: Hi @listar2000, it looks like all the tool related tests were removed. we should add equivalent ones back with the new `TinkerChatTemplateParser`?

listar2000: Yeah @luyuzhe111 you are spot on -- these are removed from `test_tinker_engine` since the original functions we tested are no longer used. These utility functions are now a part of the `TinkerChatTemplateParser`. I will add these tests back to the corresponding `test_tinker_parser.py` file.

tests/integration/test_tinker_gateway_integration.py
```diff
@@ -34,7 +34,6 @@ def create_tinker_engine():
         max_response_length=128,
         max_model_length=2048,
         sampling_params={"train": {"temperature": 0.0}, "val": {"temperature": 0.0}},
-        bypass_render_with_parser=True,
         disable_thinking=True,
```
luyuzhe111, line 38: this test will fail since `disable_thinking: bool = False` has been removed. why was it removed?

listar2000: You mean the `bypass_render_with_parser` flag or sth else? I think the removal of `bypass_render_with_parser` is the only change here.

listar2000: Oh I see the point you've made. Yeah I need to go back and check the argument list of the `TinkerEngine`.

rllm/parser/tinker_parser.py
```diff
@@ -0,0 +1,400 @@
...
+            if isinstance(chunk, tinker.EncodedTextChunk):
+                tokens.extend(chunk.tokens)
+            else:
+                # ImageChunk or other non-token chunk - use length as placeholder
+                # This path is for VL models; decode will produce placeholder tokens
+                tokens.extend([0] * chunk.length)
+
+        return tokens
+
+    def _prepare_messages(self, messages: list[dict], tools: list[Tool | dict] | None = None) -> list[dict]:
```
luyuzhe111: it looks like these functions assume the tool format is rllm format instead of openai format. how does it work with agentflow & gateway, where the messages and tool calls are in openai format?

luyuzhe111:
@listar2000 great work with `TinkerChatTemplateParser` and the refactoring! I would suggest breaking it into multiple PRs tho for easier review & history tracking. For example, consider:

- Base RolloutEngine enrichment + engine migration (move from experimental to canonical, add the new abstract methods, backward-compat shim) 
- TinkerChatTemplateParser + TinkerEngine refactor (the parser, removing bypass_render_with_parser)
- VerlEngine refactor (is_validation rename, new TITO methods, stop_reason mapping)

rllm/parser/tinker_parser.py
listar2000: @luyuzhe111  You make a good point (I think this is where this [linear issue](https://linear.app/rllm-project/issue/RLL-153/define-the-rllm-message-and-tracing-format) was raised) -- after all the rLLM lacks a well-defined messages format standard (it's mostly like OpenAI messages format, but also has its own `ToolCall` classes). I guess for here what I will need to do is (at least temporarily) detect for both rLLM and OpenAI format, and then process differently to Tinker.

listar2000:
@luyuzhe111 Agree. But for now I think 2 PRs separating the purpose seems reasonable.

I will try to get the migration done first -- i.e. making sure there will be no "dual" rollout engine paths. Meanwhile the `TinkerEngine` stays in the original form.

Meanwhile I can work on another PR making sure the `TinkerChatTemplateParser` works well for both message formats.

listar2000:
@luyuzhe111 This PR currently only contains the 1st part (i.e. the migration of rollout engines outside `experimental/`). I think it should not block the agent flow or gateway (which are already using the experimental rollout engine anyway?).

As for the Verl engine-related changes: the renaming has been manually made sure consistent, and the new TITO methods are not novel things, but a breakdown of the originally super long `get_model_response` method (`messages -> ModelOutput`) into more semantically clear stages (`messages -> token inputs -> token outputs -> ModelOutput`).

Will also be running some real tests to see how the older trainers work with this new Verl Engine.

rllm/engine/rollout/verl_engine.py
```diff
@@ -77,17 +95,28 @@ async def get_model_response(self, messages: list[dict], **kwargs) -> ModelOutpu
             multi_modal_inputs = None
             prompt_ids = request_prompt_ids
 
-        prompt_length = len(prompt_ids)
-        if enforce_max_prompt_length and prompt_length > self.max_prompt_length:
-            raise TerminationEvent(TerminationReason.MAX_PROMPT_LENGTH_EXCEEDED)
+        token_output: TokenOutput = await self.get_token_output_from_token_input(token_input=request_prompt_ids, **kwargs)
```
luyuzhe111, line 98: it looks like `image_data` never made it to `**kwargs`, so not self.server_manager.generate() either? might be an old bug in the experimental path.

rllm/engine/__init__.py
```diff
@@ -31,6 +32,13 @@ def __getattr__(name):
         from .rollout.openai_engine import OpenAIEngine as _OpenAIEngine
 
         return _OpenAIEngine
+    if name == "TinkerEngine":
```
luyuzhe111, line 35: Pre-existing pattern, this PR just extends it: Two things about the `__getattr__`blocks:

1. The try/except Exception converts ImportError into AttributeError, which hides the real error. If someone writes from rllm.engine.rollout import TinkerEngine and tinker isn't installed, they'll see "module has no attribute 'TinkerEngine'" instead of "No module named 'tinker'". Since these are explicit imports (not speculative hasattr checks), letting the ImportError propagate is more helpful.

2. The as _TinkerEngine underscore alias is unnecessary — Python caches whatever __getattr__ returns on the module automatically, so the local name doesn't matter.

Could simplify to:
```python                                                                                                          
if name == "TinkerEngine":
    from .tinker_engine import TinkerEngine                                                                                 
    return TinkerEngine        
```                                                                                             
Same applies to the existing VerlEngine and OpenAIEngine blocks.

Since CC tends to pattern match a lot, would recommend putting the right pattern in this PR and fixing other cases soon too in a separate PR : )

luyuzhe111:
@listar2000 thanks for the hard work and splitting the PR!

Regarding verl related changes tho, as I mentioned above, the VerlEngine TITO refactoring (splitting get_model_response into get_token_output_from_token_input + assemble_model_output) seems to introduce a subtle image_data regression — it's computed in _get_model_response but never forwarded to get_token_output_from_token_input, so server_manager.generate() no longer receives it for VLM prompts.

Would it make sense to keep the VerlEngine TITO refactor as a follow-up PR and just do the mechanical migration here (import path change + validate → is_validation)? That way this PR stays a clean move with no behavioral changes, and we can merge it faster without needing rigorous GPU testing — it already cleans things up significantly as-is. The TITO refactor can then be  reviewed and tested properly on its own.

listar2000:
@luyuzhe111 Sounds good, thx for such detailed review! Will work on these later today.

rllm/engine/rollout/verl_engine.py
listar2000: @kylemontgomery1 cc'ing Kyle who might be more of an expert here. If I remember this correctly, the `image_data` is manually processed into `list[int]` and combined with the text prompt. 

https://github.com/rllm-org/rllm/blob/aebb38bdbd96ebb145a04a67d787e91b7560e4cb/rllm/engine/rollout/verl_engine.py#L87-L92

I think the experimental Verl engine's handle here should at least match the legacy (non-experimental) one right?

rllm/engine/__init__.py
listar2000: Fixed. Will do another PR later on reverting similar patterns across codebase.

rllm/engine/rollout/verl_engine.py
kylemontgomery1: rllm/engine/rollout/verl_engine.py in main correctly passes image_data to server_manager.generate, the experimental one does not. Merging would break multimodal support for verl in the agent workflow trainer (it may have never worked in the unified trained...)

@listar2000 I'd suggest holding off on moving the experimental verl rollout engine out of experimental/ for now; the other rollout engines seem fine to move. We have some other changes coming to the verl rollout engine to enable fully async (e.g., we will use a different server manager). Besides, the verl backend for the unified trainer is still in experimental (and rightly so imo).

listar2000: @kylemontgomery1 Sounds good and thx for the review. In this case I think I will hold off this PR until the changes to (experimental) Verl rollout engine come in (might not directly to the current branch in this PR -- but I can cherrypick those changes). 

I also agree that the Verl backend is still in experimental -- but considering that the (non-experimental) Verl workflow trainer isn't much more stable as well, it makes sense for us to keep working to make Verl backend more stable and then migrate it out soon.

listar2000: Hi @kylemontgomery1 I think it's time to revisit this PR.

Are there any major changes to the (experimental) Verl rollout engine in your incoming fully-async PR? If so I can wait for that before migrating it out of experimental. 

Also I think this PR should be merged before your next PR on rLLM gateway -- which if I understand correctly also need to touch the rollout engines slightly.

listar2000:
Close it for now since we might have a newer version for this feature.

PR #493, opened 2026-04-06
fix: unified async trainer with verl backend

yifannnwu:
## Summary

Fixes 5 bugs that prevent the async training path (`_fit_fully_async` from PR #481) from working with the verl backend. The sync path is unaffected.

### 1. `_generation_loop` batch type mismatch (`unified_trainer.py`)
`batch[0]` crashes because verl's dataloader yields dicts (with `extra_info`/`raw_prompt` keys), not indexable lists. Fix: handle `dict` and `DataProto` batch types before falling back to `batch[0]`.

### 2. `VerlEngine.__init__` missing `super().__init__()` (`verl_engine.py`)
The experimental `VerlEngine` doesn't call `super().__init__()`, so the base class `weight_version = 0` attribute is never set. The async trainer reads `rollout_engine.weight_version` for staleness tracking → `AttributeError`.

### 3. `trainer_state.episodes` not set before fwd-bwd pass (`unified_trainer.py`)
`_training_loop` never sets `trainer_state.episodes` before calling `transform_to_backend_batch`, which asserts that episodes are present. The existing assignment after the optimizer step is for visualization/logging — too late for the fwd-bwd pass.

### 4. `compute_advantages` skipped in async path (`unified_trainer.py`)
`_training_loop` calls `process_backend_batch` → `update_policy` but skips `compute_advantages`. The sync path (`_fit_sync`) has all three steps. Without advantages, `update_policy` fails with `KeyError: 'advantages'`.

### 5. `timing_dict["step"]` not populated (`unified_trainer.py`)
verl's `compute_throughout_metrics` (called by `on_batch_end`) requires `timing_raw["step"]`. The sync path uses `with simple_timer("step", timing_dict)` but the async path doesn't go through that code path.

## Test plan

- [x] Qwen3.5-0.8B, verl backend, async GRPO: 3 training steps completed successfully

### DIFF
--- a/rllm/experimental/rollout/verl_engine.py
+++ b/rllm/experimental/rollout/verl_engine.py
@@ -13,6 +13,7 @@ from rllm.workflows import TerminationEvent, TerminationReason
 
 class VerlEngine(RolloutEngine):
     def __init__(self, config: DictConfig, rollout_manager: AgentLoopManager, tokenizer: Tokenizer, processor=None, **kwargs):
+        super().__init__()
         self.config = config
 
         if config.actor_rollout_ref.rollout.name not in ["vllm", "sglang"]:
--- a/rllm/experimental/unified_trainer.py
+++ b/rllm/experimental/unified_trainer.py
@@ -543,7 +543,15 @@ class UnifiedTrainer:
                 self.agent_workflow_engine.set_training_step(trainer_state.global_step, mode="train", epoch=epoch)
 
                 for batch in train_dataloader:
-                    task = batch[0]
+                    # Extract single task from batch (train_batch_size=1).
+                    # verl dataloader yields dicts (with extra_info, raw_prompt keys),
+                    # not indexable lists.
+                    if isinstance(batch, dict):
+                        task = batch.get("extra_info", [{}])[0]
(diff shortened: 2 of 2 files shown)


yifannnwu:
cc @kylemontgomery1 for review — thanks for async path from PR #481, which is super helpful! I did a testing with verl backend + Qwen3.5.

luyuzhe111:
thanks for the fix @yifannnwu! wondering what script you used for e2e testing?

listar2000:
@yifannnwu Thx a lot for the changes!

@kylemontgomery1 -- at this point might need to figure out a good merge queue schedule for these co-existing PRs (if I remember you also have merged the #488 changes already with some additional changes to Verl engine).

kylemontgomery1:
We have some ongoing effort here: https://github.com/rllm-org/rllm/tree/verl-fully-async, but this is not ready yet.

The main change will be to decouple the backend from verl's RayPPOTrainer. 

I'd also like to drop the backend-specific advantage computation so that advantage computation can be done to the trajectory groups in the buffer, but we'd lose gae (I doubt anyone has ever used gae with rllm). Happy to keep this if there is a good reason.

PR #494, opened 2026-04-08
feat: Fix sync agent flow event loop blocking

jeffreysijuntan:
## Summary

This branch adds a new **math_tool_agent cookbook**, fixes compatibility with **verl 0.7.1** (both legacy and new `EngineWorker` paths), converts cookbook `AgentFlow`s from sync to async, and resolves a `ZeroDivisionError` that surfaced when solver-judge tinker training produced all-equal rewards within a group.

10 commits, 27 files changed (+848 / -75).

### New: `math_tool_agent` cookbook
Multi-turn math agent that trains with a calculator tool, wired for both **tinker** and **verl** backends.

- Dataset switched from GSM8K → `hendrycks_math` (train) + MATH-500 (test). GSM8K problems were too easy and completed in a single turn, which did not exercise multi-turn RL.
- Uses native OpenAI function-calling (`tools` parameter) rather than text-based `<tool_call>` XML tags, so the model's trained tool-use behavior actually activates.
- Reward is **gated on tool usage** — no reward unless the model actually calls the calculator — preventing single-turn shortcuts.
- `MAX_TURNS` raised from 5 → 8 for harder problems; `**` (power) and `%` (modulo) added to the calculator.
- New `hendrycks_math_transform` in `rllm/data/transforms.py` extracts `\boxed{...}` answers (with nested-brace handling) and falls back to `####` pattern.
- New dataset entry registered in `rllm/registry/datasets.json`.

Files: `cookbooks/math_tool_agent/{README.md,math_tool_agent.py,evaluator.py,test.py,train.py,train_tinker.sh,train_verl.sh,pyproject.toml}`

Also in `rllm/experimental/verl/verl_backend.py`:
- `set_loss_fn` call is now guarded for newer verl versions where the method is absent.
- `_execute_tasks_async` now forwards `is_validation` so post-rollout sleep is skipped during validation (preventing CUDA illegal memory access on the next generation).
- `VerlEngine.__init__` now calls `super().__init__()`.

### Sync → async cookbook AgentFlows
Converted `solver_judge_flow`, `geo3k_flow`, and `math_tool_agent` from sync `OpenAI` + `ThreadPoolExecutor` to `AsyncOpenAI` + `asyncio.gather`, so `AgentFlowFn.arun` takes the native-await path. Eliminates per-task thread spawns, sync `httpx` loopback blocking, and GIL contention under the gateway path.

As a safety net, `AgentFlowFn.arun` now dispatches sync flows to `loop.run_in_executor` so future sync flows don't block the event loop.

### Solver-judge tinker fixes
Three interlocking fixes to unblock `ZeroDivisionError` in solver-judge tinker training:

- `SolverJudgeWorkflow` now sets trajectory-level `traj.reward` alongside step-level rewards, so downstream grouping sees actual values instead of `None`.
- `tinker_workflow_trainer` reward aggregation guard now checks `is None` instead of `== 0.0` — matches `Trajectory`'s default of `None` and stops it from clobbering legitimate zero rewards.
- `TinkerAdvantageComputer` now falls back to `std=1.0` when all group rewards are identical, yielding zero advantages instead of a divide-by-zero.

Also restored `solver_judge_flow/train.sh` to match the original example configs (`Qwen3-4B-Instruct-2507`, `group_size=4`, `max_response_length=1024`).

### Misc
- `rllm-docs` added as a git submodule.
- README: Python version requirement updated.
- Ruff formatting pass applied across touched files.
- New countdown unified-trainer entry point: `examples/countdown/unified_trainer/train_countdown_unified_verl.{py,sh}`.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "rllm-docs"]
+	path = rllm-docs
+	url = https://github.com/rllm-org/rllm-docs
--- a/README.md
+++ b/README.md
@@ -150,6 +150,8 @@ Under the hood:
 - [Cut the Bill, Keep the Turns](https://agate-slipper-ef0.notion.site/Cut-the-Bill-Keep-the-Turns-Affordable-Multi-Turn-Search-RL-003f78214a4d451fb06f453d084e666c) — Affordable multi-turn search RL
 - [Experiential Reinforcement Learning](https://arxiv.org/abs/2602.13949) — Experience-reflection-consolidation loop for RL with sparse rewards
 - [V1: Unifying Generation and Self-Verification](https://arxiv.org/abs/2603.04304) — Pairwise self-verification for parallel test-time scaling
+- [TherapyGym](https://therapygym.stanford.edu/) - Evaluating and Aligning Clinical Fidelity and Safety in Therapy Chatbots
+- [SandMLE](https://arxiv.org/pdf/2604.04872) - Synthetic Sandbox for Training MLE Agents
 ## Articles & Blog Posts
 
 - [rLLM UI: Real-Time Observability Tool for Agent Training & Evaluation](https://rllm-project.com/post.html?post=rllm_ui.md) — Mar 2026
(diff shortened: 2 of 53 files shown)


luyuzhe111:
@jeffreysijuntan good work! can you also provide the training curve for verl backend? thanks!

PR #495, opened 2026-04-08
feature: harbor remote runtime

kylemontgomery1:
## Summary

Adds basic support for harbor as a remote runtime. Various changes to the gateway and engine to facilitate this. Currently we only support scaffolds from harbor that use the chat completions endpoint, working on a larger refactor of the gateway to better support more harnesses.

## Type of change

- [x] Feature
- [ ] Fix
- [ ] Docs
- [ ] Refactor
- [ ] Example / Project
- [ ] Infra / CI

## What changed

- Adds basic support for harbor as a remote runtime. 
- Various changes to the gateway layer to support harbor.
- 

## Validation

- [ ] Not run (reason below)

Validation details:
- 

- None

## Docs / examples

- [ ] Not needed
- [ ] Updated docs
- [x] Updated examples
- [x] Follow-up docs needed

## Related issues / PRs

- Fixes #
- Related to #
- Stacked on / depends on #

## Screenshots / logs

W B Chart 4_13_2026, 12_35_30 PM

W B Chart 4_20_2026, 3_07_48 PM


kylemontgomery1:
@jeffreysijuntan @luyuzhe111 Could you take a quick look? Hopefully I didn't break anything on the agentcore side.

rllm-model-gateway/src/rllm_model_gateway/middleware.py
```diff
@@ -118,18 +124,18 @@ def _mutate(self, payload: dict[str, Any], session_id: str | None = None) -> Non
...
-        if self.add_return_token_ids:
-            # extra_body is the OpenAI Python SDK convention for vendor extensions
-            extra = payload.setdefault("extra_body", {})
-            if isinstance(extra, dict) and "return_token_ids" not in extra:
-                extra["return_token_ids"] = True
-            # Also set at root for direct vLLM calls
-            if "return_token_ids" not in payload:
-                payload["return_token_ids"] = True
-        # Inject per-session sampling params (only if client didn't already set them)
+        if self.add_return_token_ids and "return_token_ids" not in payload:
```
luyuzhe111, line 127: what's the reason for this change? isn't `return_token_ids` in extra_body?

rllm-model-gateway/src/rllm_model_gateway/middleware.py
```diff
@@ -35,11 +35,17 @@ def __init__(
...
+        model: str | None = None,
     ) -> None:
+        if sampling_params_priority not in ("client", "session"):
+            raise ValueError(f"sampling_params_priority must be 'client' or 'session', got {sampling_params_priority!r}")
         self.app = app
         self.add_logprobs = add_logprobs
         self.add_return_token_ids = add_return_token_ids
         self.sessions = sessions  # SessionManager — for per-session sampling params
+        self.sampling_params_priority = sampling_params_priority
+        self.model = model
```
luyuzhe111, line 48: is this for lora?

rllm/experimental/config/rllm/base.yaml
```diff
@@ -146,21 +146,32 @@ sdk:
...
+    verifier_timeout_multiplier: null
+    agent_setup_timeout_multiplier: null
+    environment_build_timeout_multiplier: null
   session_timeout: 300.0
 
 # Gateway Configuration (model gateway for AgentFlow-based training)
 gateway:
   port: 9090
   host: null          # Auto-detects routable IP; set explicitly to override
+  public_url: null    # Public URL for remote agents (e.g. http://<tunnel-host>:9090); defaults to http://<host>:<port>
```
luyuzhe111, line 172: why is this needed

examples/agentcore_math/train_agentcore_math_tinker.sh
```diff
@@ -31,7 +31,7 @@ python -m examples.agentcore_math.train_agentcore_math_tinker \
     rllm.trainer.save_freq=100 \
     rllm.remote_runtime.enabled=true \
     rllm.remote_runtime.backend=agentcore \
-    rllm.remote_runtime.backend_config.agent_runtime_arn=$AGENTCORE_AGENT_ARN \
-    rllm.remote_runtime.backend_config.s3_bucket=$AGENTCORE_S3_BUCKET \
-    rllm.remote_runtime.backend_config.tps_limit=25 \
+    rllm.remote_runtime.agentcore.agent_runtime_arn=$AGENTCORE_AGENT_ARN \
```
luyuzhe111, line 34: how was this change tested?

rllm-model-gateway/src/rllm_model_gateway/middleware.py
```diff
@@ -35,11 +35,17 @@ def __init__(
         add_logprobs: bool = True,
         add_return_token_ids: bool = True,
         sessions: Any | None = None,
+        sampling_params_priority: str = "client",
+        model: str | None = None,
     ) -> None:
+        if sampling_params_priority not in ("client", "session"):
```
luyuzhe111, line 41: why do we need this? 

I believe right now, gateway manager should register different sampling parameters for train & val and these will be automatically attached to each request before it reaches the inference server. 

if you want even finer-grained sampling parameters, users can always pass them directly in the payload just like how they pass in different prompts? this seems like something users should handle on the application side.

examples/harbor_swe/train_harbor.sh
```diff
@@ -0,0 +1,42 @@
+set -x
```
luyuzhe111, line 1: did you run this script? if so, can you share the test result in the PR body?

luyuzhe111:
@kylemontgomery1 cool work. please make sure this integration test passes. thanks! https://github.com/rllm-org/rllm/blob/main/tests/integration/test_agentcore_runtime.py

rllm-model-gateway/src/rllm_model_gateway/middleware.py
kylemontgomery1: if we make the http request directly instead of via the OpenAI client, it should be at the top level. The OpenAI client dumps everything in extra body to the top level when sending the request.

rllm-model-gateway/src/rllm_model_gateway/middleware.py
kylemontgomery1: For harbor, the scaffold is basically a black box, so we want to be able to inject the proper parameters into the llm call (e.g., the model and and override sampling parameters). For instance, many harbor providers will do something like mode=model.split("/")[-1].

rllm/experimental/config/rllm/base.yaml
kylemontgomery1: This is to make sure the gateway is reachable by the external service (e.g., daytona). Lots of our clusters have firewall rules that block such connections. This allows us to set a public entrypoint, e.g., if you we're routing traffic through a separate server.

rllm-model-gateway/src/rllm_model_gateway/middleware.py
kylemontgomery1: `users can always pass them directly in the payload` 
For harbor, the scaffolds are basically blackboxes, and users cannot control which model or sampling params are sent in the request. Basically, we cannot assume that users have full control of the application side.

sampling_params_priority set's the priority, e.g, whether the sampling params sent by the trainer should override those sent by the sandbox when there is collisions. This needs to set to `session` (i.e., those attached to the session by the engine/trainer). In my opinion, `session` is also a much better source of truth, which is why I changed the default in the base training config. For example, if your agents sends requests with temperate=0.7, but the trainer temperate is set differently, the logprobs during training wont be correctly scaled causing training to collapse.

examples/harbor_swe/train_harbor.sh
kylemontgomery1: Ran a few steps (tinker is really expensive), will update the pr with the reward curve.

kylemontgomery1:
> @kylemontgomery1 cool work. please make sure this integration test passes. thanks! https://github.com/rllm-org/rllm/blob/main/tests/integration/test_agentcore_runtime.py

will do.

examples/agentcore_math/train_agentcore_math_tinker.sh
kylemontgomery1: will check, I can just run this example and confirm.

rllm-model-gateway/src/rllm_model_gateway/middleware.py
luyuzhe111: I see. when will we make the http request directly? can you point me to some relevant harbor documentation? thanks!

rllm/experimental/config/rllm/base.yaml
luyuzhe111: this makes sense. for agentcore we directly deploy the agent in the same vpc as the cluster so there were no such problems, but having a public url will still be very helpful so users don't have to deploy their agent to the vpc. 

how are you obtaining this public url right now?

rllm-model-gateway/src/rllm_model_gateway/middleware.py
luyuzhe111: interesting. i was not aware of harbor's constraints. again could you point me to some relevant docs. much appreciated!

rllm/experimental/config/rllm/base.yaml
kylemontgomery1: I just do a reverse ssh tunnel to some cheap hetzner vps.

listar2000:
LGTM! I will probably merge later tonight after seeing if @luyuzhe111 have any further comments. But overall thanks for the review @luyuzhe111 and I think your previous concern (integration test) has been addressed.

Regarding the docs, I think indeed we should start to craft up some documentations on the remote runtime (maybe later for rllm gateway as well). I can provide a draft first for you guys to review (since I want to use this as a chance to get familiar with these remote runtime codes as well).

PR #499, opened 2026-04-16
fix(verl): move legacy worker override to launcher before worker class selection

luyuzhe111:
### Issue
Legacy worker override happens too late. Code path:

1. add_actor_rollout_worker(config) reads config.trainer.use_legacy_worker_impl to decide which worker class to use — "disable" picks the new ActorRolloutRefWorker (from engine_workers.py), anything else picks the legacy AsyncActorRolloutRefWorker (e.g., from megatron_workers.py).
2. VerlBackend.validate_config() forces the value to "disable" — but it runs later, inside UnifiedTrainer.__init__ at line 262, after add_actor_rollout_worker has already selected the wrong worker class                                                                                                        
3. The new EngineWorker's ActorRolloutRefWorker has set_loss_fn (via `@register` decorator that gets bound to the RayWorkerGroup). The legacy AsyncActorRolloutRefWorker does not. So when init_rollout_engine called self.actor_rollout_wg.set_loss_fn(...), it crashed with AttributeError: 'RayWorkerGroup' object has no       
attribute 'set_loss_fn'

### Fix
- Forces "disable" in the launcher before add_actor_rollout_worker reads it, so the right worker class is selected from the start.
- Fixes two minor bugs in the original `logger.warning(...)` (not a f-string, no space between concatenation). 

### Test
- `examples/agentcore_math/train_agentcore_math_verl.sh` runs without the `AttributeError`.

### DIFF
--- a/rllm/experimental/verl/verl_backend.py
+++ b/rllm/experimental/verl/verl_backend.py
@@ -208,14 +208,6 @@ class VerlBackend(BackendProtocol[Iterable, DataProto], RayPPOTrainer):
         """Validate verl-specific configuration settings."""
         assert self.config.actor_rollout_ref.rollout.mode == "async", "Only async rollout mode is supported for VerlBackend"
         assert self.use_rm is False, "Reward models are not supported. Rewards should be assigned using a reward function in the workflow or environment."
-        # Enforce new EngineWorker path (TensorDict + no-padding)
-        legacy_mode = self.config.trainer.get("use_legacy_worker_impl", "auto")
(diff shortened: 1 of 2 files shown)


kylemontgomery1:
Yeah I noticed this as well and fixed it in the wip verl fully async branch https://github.com/rllm-org/rllm/blob/c58a411a2d03ab9073e7bc6b9c69ef35a0aeeeaa/rllm/experimental/verl/verl_launcher.py#L41. I'll go ahead and merge this pr.

PR #503, opened 2026-04-19
Fix rollout_probs_diff masking to respect response_mask in Verl trainers

### LINKED ISSUE #497: rollout_probs_diff metric calcu bug
### Description

in agent_ppo_train.py

`rollout_probs_diff = calculate_log_prob_diff(actor_probs, rollout_probs, response_mask_bool),`
 while mask is：
                             ```
   attention_mask = batch.batch["attention_mask"]
                                responses = batch.batch["responses"]
                                response_length = responses.size(1)
                                response_mask = attention_mask[:, -response_length:]
```
In agentic rl，response part with tool call should be masked. If using attention mask[:-response_len], the metric would be inaccurate

### Steps to Reproduce

in agent_ppo_train.py

### Error Output / Traceback

```shell

```

### rLLM Version

0.2.1post

### Training Backend

verl

### Python Version

3.9

### GPU / CUDA Version

_No response_

### vLLM Version (if applicable)

_No response_

### Training Script / Config

```shell

```

### Additional Context

_No response_

taivu1998:
## Summary

Closes #497.

This PR fixes `training/rollout_probs_diff_*` so the metric is computed only over loss-bearing response tokens, rather than every non-padding token in the response tail.

## Root Cause

The affected Verl trainer paths rebuilt the metric mask from `attention_mask[:, -response_length:]`. That mask only distinguishes padding from non-padding, so it incorrectly included response-side tokens that should be excluded from agent training metrics, such as masked tool-interaction segments.

The repo already carries the correct contract in `batch.batch["response_mask"]`: it marks the response tokens that actually participate in training. Using the looser attention-mask slice made `rollout_probs_diff` inaccurate for agentic trajectories.

## What Changed

- added a small shared helper in `rllm/experimental/verl/metrics.py` to compute rollout-vs-actor probability drift from `rollout_log_probs`, `old_log_probs`, and `response_mask`
- updated the three affected call sites to use that helper instead of rebuilding a response-tail mask from `attention_mask`
  - `rllm/trainer/verl/agent_ppo_trainer.py`
  - `rllm/trainer/verl/agent_sdk_trainer.py`
  - `rllm/experimental/verl/verl_backend.py`
- added an empty-mask guard so the metric block safely skips all-zero masks
- added a single-token guard so `training/rollout_probs_diff_std` reports `0.0` instead of `nan` when only one valid token remains after masking
- added focused regression tests for:
  - internal masked-out response tokens not affecting the metric
  - all-zero masks returning no metric keys
  - single-token masks producing a stable zero standard deviation

## Why This Design

This keeps the fix narrow and consistent:

- the masking contract now matches the rest of the Verl training path by reusing `response_mask`
- the duplicated metric logic is centralized in one small Verl-local helper instead of repeating the same math in three places
- the change does not expand scope into broader response/tokenization redesign work

## Validation

- `uv run python -m pytest tests/test_verl_metrics.py -q`
- `python -m py_compile /tmp/rllm-issue-497-worktree/rllm/experimental/verl/metrics.py /tmp/rllm-issue-497-worktree/rllm/experimental/verl/verl_backend.py /tmp/rllm-issue-497-worktree/rllm/trainer/verl/agent_ppo_trainer.py /tmp/rllm-issue-497-worktree/rllm/trainer/verl/agent_sdk_trainer.py /tmp/rllm-issue-497-worktree/tests/test_verl_metrics.py`
- `git diff --check`

## Notes

I also attempted to rerun `tests/unified_trainer/test_verl_transform.py`, but that suite requires the optional `verl` package to be installed in the temp worktree environment and currently fails during import collection without it. The changed metric path itself is covered by the focused regression tests above.

### DIFF
--- a/rllm/experimental/verl/metrics.py
+++ b/rllm/experimental/verl/metrics.py
@@ -1,0 +1,32 @@
+"""Small metric helpers shared by rLLM's Verl integrations."""
+
+from __future__ import annotations
+
+import torch
+
+
+def compute_rollout_probs_diff_metrics(
+    *,
+    rollout_log_probs: torch.Tensor,
+    actor_log_probs: torch.Tensor,
+    response_mask: torch.Tensor,
+) -> dict[str, float]:
+    """Compute rollout-vs-actor probability drift over loss-bearing response tokens only."""
+    rollout_probs = torch.exp(rollout_log_probs)
+    actor_probs = torch.exp(actor_log_probs)
+    rollout_probs_diff = torch.abs(rollout_probs - actor_probs)
+    masked_rollout_probs_diff = torch.masked_select(rollout_probs_diff, response_mask.bool())
+
+    if masked_rollout_probs_diff.numel() == 0:
+        return {}
+
+    if masked_rollout_probs_diff.numel() == 1:
+        std = 0.0
+    else:
+        std = torch.std(masked_rollout_probs_diff).detach().item()
+
+    return {
+        "training/rollout_probs_diff_max": torch.max(masked_rollout_probs_diff).detach().item(),
+        "training/rollout_probs_diff_mean": torch.mean(masked_rollout_probs_diff).detach().item(),
+        "training/rollout_probs_diff_std": std,
+    }
--- a/rllm/experimental/verl/metrics.py
+++ b/rllm/experimental/verl/metrics.py
@@ -3,12 +3,15 @@
 from __future__ import annotations
 
 import logging
+import math
 from collections.abc import Callable
 from typing import Any
 
-import torch
-
 logger = logging.getLogger(__name__)
+_EMPTY_REDUCTION_ERROR_SNIPPETS = (
+    "Expected reduction dim to be specified",
+    "input.numel() == 0",
+)
 
 
 def _load_verl_calculate_debug_metrics() -> Callable[[Any], dict[str, float]]:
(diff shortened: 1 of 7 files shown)


listar2000:
Hi @taivu1998 — this is a really nice addition and a solid fix to the issue, thanks for putting this together!

One thing to flag: in `Verl`, it looks like there’s already a [built-in metrics](https://github.com/verl-project/verl/blob/c4e1574ba51fb93a1597e960a5d42c7befa5861a/verl/utils/debug/metrics.py#L63) helper that was added recently (we didn’t quite catch up with it on the `rLLM` side yet). The [usage](https://github.com/verl-project/verl/blob/c4e1574ba51fb93a1597e960a5d42c7befa5861a/verl/trainer/ppo/ray_trainer.py#L1459) there seems quite similar to what you’ve implemented here.

Given that, I’m wondering if it might make more sense for us to rely on the backend-provided metric utilities rather than maintaining a separate implementation on our end.

That said, I definitely don’t want your effort here to go to waste — this PR is a great starting point. Would you be open to adapting it to integrate the existing `Verl` metrics helper into our codebase instead?

rllm/experimental/verl/metrics.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+def _load_verl_calculate_debug_metrics() -> Callable[[Any], dict[str, float]]:
+    """Load Verl's debug metrics helper lazily to preserve optional dependency boundaries."""
+    from verl.utils.debug.metrics import calculate_debug_metrics
+
+    return calculate_debug_metrics
+
+
+def _resolve_log_prob_mask(data: Any) -> torch.Tensor:
```
listar2000: I don't think this is actually needed? The Verl's helper function also checks for the response_mask being all False (essentially same logic as here) and also return the identical content in `_default_debug_metrics` below. No need to recompute these masks again,

rllm/experimental/verl/metrics.py
```diff
@@ -0,0 +1,63 @@
...
+    response_mask = _resolve_log_prob_mask(data)
+
+    if not response_mask.any().item():
+        logger.warning("response_mask is all False, returning default metrics")
+        return _default_debug_metrics()
+
+    metrics = _load_verl_calculate_debug_metrics()(data)
+
+    if response_mask.sum().item() == 1 and metrics.get("training/rollout_probs_diff_valid") == 1:
+        metrics["training/rollout_probs_diff_std"] = 0.0
```
listar2000: This is a good catch and improvement over Verl's native impl. Thx!

listar2000:
Hi @taivu1998 thx for the update! I will merge it once my remaining comment (recompute response mask) and the precommit (mainly unused imports) are fixed.

taivu1998:
Thanks @listar2000, I updated the PR. Could you review again?

listar2000 (approved):
thx! LGTM!

PR #504, opened 2026-04-20
fix(parser): Qwen3.5 chat template support + simple_math example

JinyanSu1:
## Summary

Three logically separate commits add Qwen3.5 support to rllm.

### 1. `fix(parser)`: support strict-alternation chat templates (Qwen3.5)

Qwen3.5's chat template raises `TemplateError` when the first message is
not `system` or `user`. `ChatTemplateParser._get_generation_prompt`
previously fed a lone assistant message to the template, which broke
`__init__` for any Qwen3.5 tokenizer. Prepend a stub user message — it
appears in both `with_prompt` and `without_prompt` so it cancels in the
slice that extracts the generation prompt (no behavior change for any
existing tokenizer).

Added two regression tests:
- A hermetic test that injects a strict-alternation Jinja template (no
  model download needed).
- An end-to-end test on `Qwen/Qwen3.5-2B`, skipped when unavailable.

### 2. `fix(parser)`: pre-existing test failures on `main`

- `test_parser_with_disable_thinking` was asserting on `\\n` (escaped
  backslashes) instead of `\n` (real newlines).
- `get_parser` no longer routed `DeepSeek-R1-Distill-Qwen-1.5B` to
  `DeepseekQwenChatTemplateParser` because newer `transformers` no
  longer wraps it as `LlamaTokenizer` (now reports as `TokenizersBackend`).
  Added `distill-qwen` to the recognition heuristic so routing keeps
  working regardless of tokenizer backend.

### 3. `example(simple_math)`: add Qwen3.5-4B FSDP2 training recipe

Qwen3.5 variant of `train_hendrycks_math.sh`. Uses FSDP2 (megatron does
not yet support Qwen3.5) with `Qwen3_5DecoderLayer` wrap policy, vLLM
`gdn_prefill_backend=triton`, and `trust_remote_code=True`. Defaults
tuned for 8x141GB H200 (`MICRO_BS=2 GPU_MEM_UTIL=0.6`); for 80GB GPUs
set `OFFLOAD=1 MICRO_BS=1`.

## Validation

`pytest tests/parser/test_chat_parser.py` — 7 passed (was 5/7 on `main`
due to the two pre-existing failures fixed in commit 2).

End-to-end Qwen3.5-4B GRPO on Hendrycks MATH runs cleanly with the new
recipe — wandb link to be added after the run completes.

### DIFF
--- a/examples/simple_math/train_hendrycks_math_qwen3_5.sh
+++ b/examples/simple_math/train_hendrycks_math_qwen3_5.sh
@@ -1,0 +1,117 @@
+#!/usr/bin/env bash
+# Qwen3.5 GRPO on Hendrycks MATH. Variant of train_hendrycks_math.sh.
+#
+# Qwen3.5 requires: actor.strategy=fsdp2 (megatron not yet supported),
+# Qwen3_5DecoderLayer wrap policy, vLLM gdn_prefill_backend=triton, and
+# trust_remote_code=True.
+#
+# Defaults are tuned for 8x141GB H200; for 80GB GPUs set OFFLOAD=1
+# MICRO_BS=1.
+#
+# Usage:
+#   python3 examples/simple_math/prepare_math_dataset.py
+#   bash examples/simple_math/train_hendrycks_math_qwen3_5.sh
+
+set -ex
+
+export VLLM_ATTENTION_BACKEND=FLASH_ATTN
+export PYTORCH_CUDA_ALLOC_CONF="expandable_segments:False"
+export VLLM_USE_V1=1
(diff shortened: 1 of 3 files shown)


kylemontgomery1:
lgtm! Thank you for contributing this.

JinyanSu1:
I just finished testing the scripts, the training curve looks good. 

Do I also need to provide guidelines for conda environment? (I don't remember if the current rllm environment already supports Qwen3.5

Screenshot 2026-04-20 at 10 59 38 PM

PR #506, opened 2026-04-23
fix(verl): pad multi-turn batch to lcm of dp_size and ppo mini-batch

luyuzhe111:
## Summary
Multi-turn rollouts produce a variable number of step rows, so the post-rollout batch size no longer matches world_size alone. Two independent constraints must hold:

1. chunk(dp_size) in the worker-group dispatcher — requires total_rows % dp_size == 0.
2. make_iterator inside train_mini_batch — requires per_rank_rows % (ppo_mini_batch_size * rollout.n / dp_size) == 0, i.e. globally, total_rows % (ppo_mini_batch_size * rollout.n) == 0.

Pad to lcm(dp_size, ppo_mini_batch_size * rollout.n). Query dp_size via dispatch_info (mirroring RayPPOTrainer._get_dp_size) instead of worker_group.world_size, which over-pads by TP * PP * CP * SP.

Also adjusted the agentcore training script to work around a couple of verl 0.7.1 backend errors.

## Type of change

- [ ] Feature
- [x] Fix
- [ ] Docs
- [ ] Refactor
- [ ] Example / Project
- [ ] Infra / CI

## Test
`bash examples/agentcore_math/train_agentcore_math_verl.sh` finished without assertion errors caused by the two requirements mentioned in Summary. The collapse problem still remains. Fix will be provided in a follow-up PR. 

Screenshot 2026-04-23 at 1 11 42 AM

### DIFF
--- a/examples/agentcore_math/train_agentcore_math_verl.sh
+++ b/examples/agentcore_math/train_agentcore_math_verl.sh
@@ -5,11 +5,15 @@ set -eux
 # Load environment variables (AGENTCORE_AGENT_ARN, AGENTCORE_S3_BUCKET)
 set -a && source .env && set +a
 
-export CUDA_DEVICE_MAX_CONNECTIONS=1
 export VLLM_ALLREDUCE_USE_SYMM_MEM=0
+
+# Clean up stale ZMQ sockets from previous runs
+rm -f /tmp/rl-colocate-zmq-*.sock
 
 MODEL_PATH=Qwen/Qwen3-4B-Instruct-2507
 
+# actor_rollout_ref.actor.megatron.param_offload=True is required for
+# verl 0.7.1 to work around device mismatch errors caused by unconditional param offloading
 python -m examples.agentcore_math.train_agentcore_math_verl \
     rllm/backend=verl \
     model_engine=megatron \
@@ -37,6 +41,7 @@ python -m examples.agentcore_math.train_agentcore_math_verl \
     actor_rollout_ref.actor.megatron.use_dist_checkpointing=False \
     actor_rollout_ref.actor.megatron.use_mbridge=True \
     actor_rollout_ref.actor.megatron.vanilla_mbridge=False \
+    actor_rollout_ref.actor.megatron.param_offload=True \
     actor_rollout_ref.actor.loss_agg_mode=token-mean \
     actor_rollout_ref.model.use_remove_padding=True \
     actor_rollout_ref.rollout.tensor_model_parallel_size=1 \
--- a/rllm/experimental/verl/verl_backend.py
+++ b/rllm/experimental/verl/verl_backend.py
@@ -288,38 +288,61 @@ class VerlBackend(BackendProtocol[Iterable, DataProto], RayPPOTrainer):
         batch = batch.select_idxs(non_pad_step_indices)  # This batch only has non_pad steps
         return batch
 
-    def _get_dp_world_size(self) -> int | None:
-        """Compute the LCM of all worker group world sizes for DP splitting."""
+    def _get_dp_size(self, worker_group, mesh_name: str) -> int:
+        """Query actual DP size for a worker group mesh via dispatch info.
 
-        world_sizes = []
+        Mirrors ``RayPPOTrainer._get_dp_size``: the dispatch-info mapping
+        assigns a dp_rank to each global worker rank, so dp_size is
+        ``max(dp_rank_mapping) + 1``.
+        """
+        if mesh_name not in worker_group._dispatch_info:
+            worker_group._dispatch_info[mesh_name] = worker_group._query_dispatch_info(mesh_name)
+        return max(worker_group._dispatch_info[mesh_name]) + 1
+
+    def _get_aggregate_dp_size(self) -> int | None:
+        """Compute the LCM of DP sizes across all active worker-group meshes.
+
+        Uses the per-mesh dp_size queried from dispatch info, not the
+        worker_group ``world_size``.  They differ when TP/PP/CP > 1.
+        """
+        dp_sizes: list[int] = []
         if self.use_critic and self.critic_wg.world_size != 0:
-            world_sizes.append(self.critic_wg.world_size)
+            dp_sizes.append(self._get_dp_size(self.critic_wg, "train"))
         if self.use_reference_policy and self.ref_policy_wg.world_size != 0:
-            world_sizes.append(self.ref_policy_wg.world_size)
+            dp_sizes.append(self._get_dp_size(self.ref_policy_wg, "ref"))
         if self.use_rm and self.rm_wg.world_size != 0:
-            world_sizes.append(self.rm_wg.world_size)
+            dp_sizes.append(self._get_dp_size(self.rm_wg, "train"))
         if self.hybrid_engine:
             if self.actor_rollout_wg.world_size != 0:
-                world_sizes.append(self.actor_rollout_wg.world_size)
+                dp_sizes.append(self._get_dp_size(self.actor_rollout_wg, "actor"))
         else:
             if hasattr(self, "actor_wg") and self.actor_wg.world_size != 0:
-                world_sizes.append(self.actor_wg.world_size)
+                dp_sizes.append(self._get_dp_size(self.actor_wg, "actor"))
             if hasattr(self, "rollout_wg") and self.rollout_wg.world_size != 0:
-                world_sizes.append(self.rollout_wg.world_size)
-        if not world_sizes:
+                dp_sizes.append(self._get_dp_size(self.rollout_wg, "actor"))
+        if not dp_sizes:
             return None
-        return reduce(math.lcm, world_sizes)
+        return reduce(math.lcm, dp_sizes)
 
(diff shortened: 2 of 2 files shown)


listar2000:
Hi @luyuzhe111 great work. Taking a look now!

rllm/experimental/verl/verl_backend.py
```diff
@@ -288,38 +288,61 @@ def _remove_padding(self, batch: DataProto) -> DataProto:
...
+        dp_sizes: list[int] = []
         if self.use_critic and self.critic_wg.world_size != 0:
-            world_sizes.append(self.critic_wg.world_size)
+            dp_sizes.append(self._get_dp_size(self.critic_wg, "train"))
         if self.use_reference_policy and self.ref_policy_wg.world_size != 0:
-            world_sizes.append(self.ref_policy_wg.world_size)
+            dp_sizes.append(self._get_dp_size(self.ref_policy_wg, "ref"))
         if self.use_rm and self.rm_wg.world_size != 0:
-            world_sizes.append(self.rm_wg.world_size)
+            dp_sizes.append(self._get_dp_size(self.rm_wg, "train"))
```
listar2000, line 314: Does the second argument (aka role name) for `_get_dp_size` matter? If so I think we should put `rm` instead of `train` here? Similar for other ones.

rllm/experimental/verl/verl_backend.py
```diff
@@ -288,38 +288,61 @@ def _remove_padding(self, batch: DataProto) -> DataProto:
...
         from verl.protocol import pad_dataproto_to_divisor
 
-        world_size = self._get_dp_world_size()
-        if world_size is None:
+        dp_size = self._get_aggregate_dp_size()
+        if dp_size is None:
             return batch
 
+        # Also account for mini-batch divisibility required by make_iterator
+        # in the new EngineWorker path. train_mini_batch needs:
```
listar2000, line 335: Yeah this is a requirement for `EngineWorker` path (i.e. `self.use_legacy_worker_impl == "disable"`)
But to make this clearer (i.e. readers know where this comes from), we probably should refer to https://github.com/verl-project/verl/blob/release/v0.7.1/verl/trainer/ppo/ray_trainer.py#L1174 in the comment.

listar2000:
Just finished the review and I think these are very valid fixes! 

Also used this chance to more carefully go through `Verl`'s new EngineWorker path -- and can confirm that these batch sizes math are correct.

Only thing (unrelated to the code changes but your summary above): I think `worker_group.world_size` should equate its dp_size as they explicitly check this in places like [this](https://github.com/verl-project/verl/blob/fc6e003cd16bd78d4234995444f69bc20492118e/verl/single_controller/base/decorator.py#L273)? But nevertheless I agree that we should use `_get_dp_size`.

examples/agentcore_math/train_agentcore_math_verl.sh
```diff
@@ -5,11 +5,15 @@ set -eux
...
-export CUDA_DEVICE_MAX_CONNECTIONS=1
 export VLLM_ALLREDUCE_USE_SYMM_MEM=0
 
+# Clean up stale ZMQ sockets from previous runs
+rm -f /tmp/rl-colocate-zmq-*.sock
+
 MODEL_PATH=Qwen/Qwen3-4B-Instruct-2507
 
+# actor_rollout_ref.actor.megatron.param_offload=True is required for
+# verl 0.7.1 to work around device mismatch errors caused by unconditional param offloading
```
lyzustc, line 16: Currently this verl commit [https://github.com/verl-project/verl/commit/7db6aadd46abc421fc57b801988b7d9f9f0ae8c8](https://github.com/verl-project/verl/commit/7db6aadd46abc421fc57b801988b7d9f9f0ae8c8) works for current rllm code and does not need param offload.

luyuzhe111: yea also found this post-release commit, but worried they introduce additional bugs. since the workaround is simple enough we can easily remove them after the next release?

luyuzhe111:
@listar2000 thanks for the detailed review! will clarify things and make necessary updates in a separate PR if needed.

rllm/experimental/verl/verl_backend.py
luyuzhe111: Good question! Yep they matter. I just took another look. Concretely:

The mesh names target the new EngineWorker path:
                                                                                                                                
- actor_rollout_wg → engine_workers.ActorRolloutRefWorker, registers "actor" and "ref" (engine_workers.py:501, 542)
- critic_wg → CriticWorker gets aliased to TrainingWorker in main_ppo.py:187-190 on this path, so the registered mesh is "train" (engine_workers.py:124), not "critic". The "critic" name only applies to the legacy fsdp_workers.CriticWorker, which this launcher now bypasses.                                                                      
- ref_policy_wg → resolves to the same ActorRolloutRefWorker (ray_trainer.py:798 / ref_in_actor), so "ref".                   
                                                                                                                                
For rm_wg: I just realized the branch is unreachable as configured — VerlBackend.validate_config asserts use_rm is False, and the new-path launcher doesn't register an RM worker class either (main_ppo.py:243-253 only adds it to the resource pool). The actor_wg/rollout_wg else-branch is also dead: hybrid_engine is asserted True in RayPPOTrainer.__init__, and self.actor_wg /   
  self.rollout_wg also aren't assigned anywhere in verl 0.7.1, so the hasattr guards would always return False.

Happy to do a follow-up that (a) adds a docstring clarifying the mesh-name mapping and (b) drops the unreachable rm_wg / non-hybrid branches. The code can be simplified to:

```python
def _get_aggregate_dp_size(self) -> int | None:
      """Compute the LCM of DP sizes across all active worker-group meshes.                                                     
                                                                                                                                
      Mesh names target the new EngineWorker path (verl_launcher pins                                                           
      ``use_legacy_worker_impl='disable'``):                                                                                    
        - actor_rollout_wg -> ``engine_workers.ActorRolloutRefWorker``                                                          
          registers ``"actor"`` and ``"ref"``.                                                                                  
        - critic_wg       -> ``CriticWorker`` is aliased to ``TrainingWorker``                                                  
          in ``main_ppo.add_critic_worker``, which registers ``"train"``.                                                       
        - ref_policy_wg   -> same ``ActorRolloutRefWorker`` as actor_rollout_wg,                                                
          so the registered mesh is ``"ref"``.                                                                                                                                          
      """                                                                                                                       
      dp_sizes: list[int] = []                                                                                                  
      if self.use_critic and self.critic_wg.world_size != 0:                                                                    
          dp_sizes.append(self._get_dp_size(self.critic_wg, "train"))                                                           
      if self.use_reference_policy and self.ref_policy_wg.world_size != 0:                                                      
          dp_sizes.append(self._get_dp_size(self.ref_policy_wg, "ref"))                                                         
      if self.actor_rollout_wg.world_size != 0:                                                                                 
          dp_sizes.append(self._get_dp_size(self.actor_rollout_wg, "actor"))                                                    
      if not dp_sizes:                                                                                                          
          return None                                                                                                           
      return reduce(math.lcm, dp_sizes)
```

rllm/experimental/verl/verl_backend.py
luyuzhe111: Good call will add!

luyuzhe111:
@listar2000 regarding the last comment. thanks for the careful second pass! to clarify:

Line 273 asserts: assert len(worker_group._dispatch_info[mesh_name]) == worker_group.world_size. 
                                                                                                                                
That's len(mapping) == world_size, i.e. "every global rank has exactly one dp_rank entry". It's not dp_size == world_size. The next line makes that clear:                                                                                                  

dp_size = max(dp_rank_mapping) + 1                                                                                            
                                                                                                                                
For world_size=4, TP=2, the mapping is [0, 0, 1, 1]: length 4 (= world_size), but dp_size = 2. The assertion holds and dp_size != world_size simultaneously.                                                                                                
                                                                                                                                
So the original premise for switching to _get_dp_size (world_size over-pads by TP×PP×CP×SP) is correct. The assertion they pointed to is enforcing a different invariant — it's a sanity check that _query_dispatch_info returned a mapping covering all workers.

luyuzhe111:
@listar2000 I have submitted https://github.com/rllm-org/rllm/pull/509 to address your comments. appreciate a review. thanks!

listar2000:
> @listar2000 regarding the last comment. thanks for the careful second pass! to clarify:
> 
> Line 273 asserts: assert len(worker_group._dispatch_info[mesh_name]) == worker_group.world_size.
> 
> That's len(mapping) == world_size, i.e. "every global rank has exactly one dp_rank entry". It's not dp_size == world_size. The next line makes that clear:
> 
> dp_size = max(dp_rank_mapping) + 1
> 
> For world_size=4, TP=2, the mapping is [0, 0, 1, 1]: length 4 (= world_size), but dp_size = 2. The assertion holds and dp_size != world_size simultaneously.
> 
> So the original premise for switching to _get_dp_size (world_size over-pads by TP×PP×CP×SP) is correct. The assertion they pointed to is enforcing a different invariant — it's a sanity check that _query_dispatch_info returned a mapping covering all workers.

This is great! Thx for the clarification.

PR #507, opened 2026-04-23
feat: fully-async training with verl

kylemontgomery1:
## Summary

Adds support for fully-async training with verl backend through the unified trainer. Also fixes misc issues with fully async

## Type of change

- [x] Feature
- [ ] Fix
- [ ] Docs
- [ ] Refactor
- [ ] Example / Project
- [ ] Infra / CI

## What changed

- 
- 
- 

## Validation

- [ ] Not run (reason below)

Validation details:
- 

- None

## Docs / examples

- [ ] Not needed
- [ ] Updated docs
- [ ] Updated examples
- [ ] Follow-up docs needed

## Related issues / PRs

- Fixes #
- Related to #
- Stacked on / depends on #


rllm/trainer/verl/ray_runtime_env.py
```diff
@@ -70,6 +73,33 @@ def _get_forwarded_env_vars():
 
 
 def get_ppo_ray_runtime_env():
+    """Build the runtime_env to pass to ray.init().
```
listar2000, line 76: One doc-related comment: I think we will definitely need to update the doc to make it clear about the ray overriding priority, but then I realize that our new (mintlify) doc is lacking this page entirely: https://rllm-project.readthedocs.io/en/latest/api/trainer/ray_runtime_env/ from old doc.

Maybe let's address all these doc changes together in a later PR (we will definitely need a dedicated page to fully async training + some examples).

rllm/trainer/tinker/tinker_backend.py
```diff
@@ -418,8 +418,8 @@ async def on_train_end(self, trainer_state: TrainerState) -> None:
         """Called at the end of training."""
         assert self.policy_trainer is not None, "policy_trainer is not initialized"
 
-        # Save final checkpoint if we didn't just save it in the last batch
-        if trainer_state.global_step % self.full_config.rllm.trainer.save_freq != 0:
+        save_freq = self.full_config.rllm.trainer.save_freq
+        if save_freq <= 0 or trainer_state.global_step % save_freq != 0:
```
listar2000, line 422: I think having `save_freq <= 0` simply means we don't want to save at all? For `Tinker` I think you get charged for storage as well so sometimes I just don't want to save.

rllm/experimental/common/config.py
```diff
@@ -146,8 +172,10 @@ class RolloutCorrectionConfig:
     """
 
     tis_mode: str | None = None
-    bypass_mode: bool = True
```
luyuzhe111, line 149: nice fix

rllm/experimental/config/rllm/backend/verl.yaml
```diff
@@ -1,65 +1,37 @@
...
+# rllm.* config values are propagated to verl-native keys at runtime by
+# propagate_rllm_to_verl_config() in rllm/experimental/verl/utils.py.
 
 defaults:
   - /ppo_trainer
   - _self_
 
 actor_rollout_ref:
+  actor:
+    use_rollout_log_probs: True
```
luyuzhe111, line 14: qq: what's the connection between this and bypass mode.

rllm/experimental/verl/utils.py
```diff
@@ -0,0 +1,347 @@
...
+        if os.path.exists(dataloader_path):
+            state_dict = torch.load(dataloader_path, weights_only=False)
+            train_dataloader.load_state_dict(state_dict)
+        else:
+            print(f"Warning: No dataloader state found at {dataloader_path}, will start from scratch")
+
+    return global_steps
+
+
+def balance_batch(
```
luyuzhe111, line 205: how is this different from the current balance batch function?

rllm/experimental/config/rllm/backend/verl.yaml
kylemontgomery1: basically we always want rollout logprobs, even just for logging purposes. with bypass mode or tis/rs correction enabled, this must be true.

rllm/experimental/verl/utils.py
kylemontgomery1: currently, balance batch is a class method of RayPPOTrainer. We no longer inherit from RayPPOTrainer after this pr, so we copy it to rllm.

rllm/trainer/tinker/tinker_backend.py
kylemontgomery1: this only hits once training is finished. let me check how verl/tinker handle this. maybe it's better to have some `save_at_train_end` config flag that defaults to true.

pyproject.toml
```diff
@@ -150,7 +150,7 @@ code-tools = [
 ]
 
 harbor = [
-    "harbor==0.3.0",
+    "harbor==0.3.0 ; python_version >= '3.12'",
```
kylemontgomery1, line 153: this should be a seperate pr.

rllm/trainer/tinker/tinker_backend.py
listar2000: TODO: add a `do_save` flag for explicit handling or keep the current `<= 0 -> no save` behavior.

listar2000:
@kylemontgomery1 I've just updated the `step_id` logic following the `main` branch (both Sijun and you have done similar corrections, let's go with what's been merged already). Some new runs might be needed.

PR #509, opened 2026-04-24
refactor(verl): document EngineWorker mesh names and drop dead branches

luyuzhe111:
## Summary

Follow-up to #506. Clarifies why the hardcoded mesh names in `_get_aggregate_dp_size` are correct for the new EngineWorker path and removes dead branches that implied support for code paths rllm doesn't exercise.                                                                                            
                                                                                                               
- Docstring maps each worker group to the class that registers its                                           
  mesh (`engine_workers.TrainingWorker` -> `"train"`,                                                        
  `ActorRolloutRefWorker` -> `"actor"`/`"ref"`).                                                             
- Drop `rm_wg` branch: `validate_config` asserts `use_rm is False`                                           
  and the new-path launcher doesn't register an RM worker.                                                   
- Drop non-hybrid `actor_wg`/`rollout_wg` branch: `hybrid_engine` is                                         
  asserted True in `RayPPOTrainer.__init__` and neither attribute is                                         
  assigned anywhere in verl 0.7.1.                                                                           
- Reference `RayPPOTrainer._update_actor` for the mini-batch divisor.                                        
                                                                                                               
## Type of change                                                                                            
- [ ] Feature                                                                                                
- [x] Refactor                                                                                               
- [ ] Fix
- [ ] Docs                                                                                                   
                  
## Test                                                                                                      
`bash examples/agentcore_math/train_agentcore_math_verl.sh` runs without assertion errors. No behavior change — same mesh names are passed to `_get_dp_size`, just fewer unreachable branches.

PR #511, opened 2026-04-26
[feat] Fix training collapse for AgentFlow with verl backend

jeffreysijuntan:
Merges `dev-multiturn` → `main`. Five new commits on top of three already-merged PRs (#503, #508, #509).

## Summary

- **Unblocks AgentFlow training under the verl backend** with three correctness fixes: per-step ID collisions in advantage scatter, trailing-malformed traces failing whole rollouts, and contaminated traces leaking from prior retry attempts.
- **Adds a `calculate_debug_metrics_compat` shim** in `rllm/experimental/verl/metrics.py` that backfills the legacy empty-mask behavior verl removed, and routes the four trainers (`agent_ppo_trainer`, `agent_sdk_trainer`, `agent_workflow_trainer`, `fully_async_trainer`) plus `verl_backend` through it. Also covered by a new `tests/test_verl_metrics.py`.
- **Hardens the gateway** so `/admin/flush` drains in-flight `_safe_store` tasks, `adelete_session` flushes before deletion, and noisy admin/health paths stop spamming `uvicorn.access`.
- **Registers `deepscaler_math`** (`agentica-org/DeepScaleR-Preview-Dataset`, ~40K AIME/AMC/Omni-MATH problems) in the dataset catalog and switches `math_tool_agent` to it to reduce GRPO uniform-group rate.
- **Rewrites the math_tool_agent calculator** with an AST whitelist (sqrt, log, trig, factorial, comb/binom, etc.) so models can write `sqrt(64 + 225)` or `binom(10, 3)` without being rejected as "invalid characters".
- **Restructures cookbook scripts** for `geo3k`, `solver_judge_flow`, and `math_tool_agent` into matched `train_tinker.sh` / `train_verl.sh` pairs, and ports the verl-stable hyperparameters (`lr 1e-6`, `loss_agg_mode=seq-mean-token-mean`, asymmetric clip 0.2/0.28) into the verl scripts.

## Why

Multi-turn AgentFlow training under verl was crashing within 3–5 steps with `EnrichMismatchError` and silently producing wrong advantages even when it didn't crash. Three issues compounded:

1. `_process_trajectory` keyed `step_id` as `f"{task_id}_{trajectory.name}_step{step_idx}"`. With multiple rollouts per task, or multiple same-named trajectories per episode (solver_judge with N solver trajectories), the dict in `update_dataproto_with_advantages` saw colliding writes — only the last advantage survived, then got scattered to **all** colliding rows. Silent GRPO destruction.
2. The strict `EnrichMismatchError` rejected any rollout where traces and agent steps disagreed. In practice, multi-turn agents commonly produce one trailing malformed trace (vLLM returns an empty body when prompt+max_tokens hits `max_model_len`, or disconnects mid-stream during weight sync; the OpenAI client raises; the agent breaks without recording its Nth Step). With `MAX_TURNS=8`, the failure rate compounded enough to exhaust retries.
3. On retry, the gateway still held the failed attempt's traces, and the positional match in `_enrich_episode` mixed stale traces with new steps.

End-to-end verification on math_tool_agent with the new config: training reward climbs from 0.27 baseline to ~0.40–0.50 by step 20; math500 val pass@1 = 0.671 @ step 10, 0.686 @ step 20.

## Files of note

| Area | Files |
|---|---|
| AgentFlow/verl correctness | `rllm/experimental/verl/transform.py`, `rllm/experimental/engine/agent_flow_engine.py`, `rllm/experimental/engine/gateway_manager.py` |
| Gateway robustness | `rllm-model-gateway/src/rllm_model_gateway/server.py` |
| Dataset / agent | `rllm/registry/datasets.json`, `rllm/data/transforms.py`, `cookbooks/math_tool_agent/math_tool_agent.py`, `cookbooks/math_tool_agent/test.py`, `cookbooks/math_tool_agent/train.py` |
| Cookbook scripts | `cookbooks/{geo3k,math_tool_agent,solver_judge_flow}/train_tinker.sh`, `cookbooks/{geo3k,math_tool_agent,solver_judge_flow}/train_verl.sh`, READMEs |

## Behavior changes worth flagging to reviewers

- **`step_id` format change** (`{trajectory.uid}_step{idx}` instead of `{task_id}_{name}_step{idx}`). External code that re-keyed off the old format will break. `trajectory_ids` is *intentionally unchanged* — verl's `compute_advantage` and the broadcast helper rely on those collisions for grouping.
- **Trailing malformed traces are now silently dropped** when `traces > agent_steps` and the extras have empty `prompt_ids`/`completion_ids`. Mid-sequence empties and `traces < agent_steps` still raise `EnrichMismatchError`. A warning log is emitted on every drop.
- **`adelete_session` now flushes the gateway client before deletion**; the sync `delete_session` was removed (no callers).

## Test plan

Previous training run for multi-turn RL (math-tool-agent) crashes to 0 reward within a few steps (due to the step_id name collision), the new training run is now stable after 20 steps, and match the curve of Tinker trainer:

Screenshot 2026-04-25 at 10 06 36 PM

Solver-judge flow training is also stable:
Screenshot 2026-04-25 at 10 10 30 PM

- [x] math_tool_agent verl smoke: `bash cookbooks/math_tool_agent/train_verl.sh` — confirm reward climbs past 0.40 by step ~20 and math500 val pass@1 lands ≥0.65
- [x] solver_judge_flow verl smoke: `bash cookbooks/solver_judge_flow/train_verl.sh` — confirm solver does not collapse from 0.6 → <0.1 in one update (the failure mode the old config showed)


listar2000:
Great adds-on, taking a look right now @jeffreysijuntan

cookbooks/math_tool_agent/train_verl.sh
```diff
@@ -27,18 +27,24 @@ python -u train.py \
     +actor_rollout_ref.model.lora.alpha=32 \
```
listar2000, line 27: Why do we need these extra "+" for lora related Verl configs?

rllm/experimental/engine/agent_flow_engine.py
```diff
@@ -272,11 +289,73 @@ def _enrich_episode(
         """
         if not traces:
             logger.warning("[%s] No traces found — returning episode without token data", uid)
-            return episode
+            # Coerce to the training Trajectory/Episode subclass so downstream
+            # pydantic validators (e.g. TrajectoryGroup.trajectories) accept
+            # instances produced by agents that imported from rllm.types.
+            return Episode(
```
listar2000, line 295: Still find that we have two types of the same name (`Episode`, `Step`, etc.) a bit confusing -- any idea on how to deal with more systematically? Like can we only preserve a single `Episode` base model class, but depending on which fields are populated we can write some functions to determine it's `is_train_valid` (for training) or `is_valid` (for SDK)?

rllm/experimental/engine/agent_flow_engine.py
```diff
@@ -272,11 +289,73 @@ def _enrich_episode(
...
+        # final LLM call was captured by the gateway (request/response made it
+        # through the proxy) but vLLM returned an empty body — typically because
+        # prompt + max_tokens hit max_model_len, or vLLM disconnected mid-stream
+        # during weight sync. The OpenAI client raises on the empty response,
+        # the agent catches it and breaks *without* recording a Step. Result:
+        # traces = N+1 with the (N+1)-th malformed; agent_steps = N with all N
+        # backed by valid leading traces. Discard the abandoned trailing traces
+        # and proceed — the prefix carries real training signal. Strict failure
+        # here would burn the entire rollout, and at large MAX_TURNS the failure
+        # rate compounds enough to exhaust retries and crash training.
```
listar2000: Small question: are these multi-line comments written by human or Claude generated? If the later case, feel like we should try to ether simplify them or put "design rationale" (i.e. "why") related details into doc, while leaving only "what" for the implementation itself.

rllm/experimental/engine/agent_flow_engine.py
jeffreysijuntan: Yes, I think this is an issue with migrating old Episode/Trajectory/Step object to the new one in `/types`. Right now the old Episode class become subclass of the new Episode class.  For backward compatibility, a temporary workaround is just to keep the existing old Episodes. Merging these two requires some dedicated efforts. I might have a separate PR in the future to merge these two types into one.

rllm/experimental/verl/transform.py
```diff
@@ -245,7 +245,13 @@ def _process_trajectory(trajectory: Trajectory, task_id: str, accumulated: Accum
...
         step_reward = step.reward
         multi_modal_inputs = step.model_output.multi_modal_inputs or {}
-        step_id = f"{trajectory_id}_step{step_idx}"
+        # step_id must uniquely identify this row across the whole batch so
+        # update_dataproto_with_advantages can scatter per-trajectory advantages
+        # back without collisions. trajectory_id alone is f"{task_id}_{name}",
+        # which collides across rollouts of the same task and across multiple
+        # same-named trajectories within one episode (e.g. solver-judge with N
+        # solver trajectories). trajectory.uid is a per-Trajectory UUID4.
+        step_id = f"{trajectory.uid}_step{step_idx}"
```
listar2000, line 254: This is a valid fix -- Kyle used the tuple (episode_id, step_id) for unique identification, but I think the one here is better (it can be hard to get episode_id for trajectory groups).

listar2000:
@jeffreysijuntan Hi JT thx again for the important fix! Just reviewed everything (even though the gateway part is not too familiar to me) and LGTM!

listar2000 (approved):

rllm/experimental/verl/transform.py
listar2000: Again I think this raises the question of our usage of `trajectory_id` in other places -- I will take a systematic look and also add KL support in rLLM native path.

neuralhippo:
FYI, this issue still persists when training multi-turn tool agents with agentflow using `Qwen/Qwen3-1.7B` and the `hermes` parser:

```python
File "../rllm/rllm/engine/agentflow_engine.py", line 179, in enrich_episode_with_traces
    raise EnrichMismatchError(f"[{uid}] enrich mismatch: traces={len(training_steps)} agent_steps={n_agent_steps} empty_prompt_ids={empty_prompt} empty_completion_ids={empty_compl}")
rllm.engine.agentflow_engine.EnrichMismatchError: [e3c223cb-511d-4bdc-8d76-3e89377dee1a:0] enrich mismatch: traces=5 agent_steps=2 empty_prompt_ids=1 empty_completion_ids=1
```

PR #521, opened 2026-04-30
fix(ray): merge ray job submit runtime_env with rllm defaults

kylemontgomery1:
Update the runtime_env we hand to `ray.init()` so it composes cleanly with `ray job submit --runtime-env-json=...` instead of colliding on shared keys (`NCCL_DEBUG`, `VLLM_LOGGING_LEVEL`, `working_dir`, ...).

## Summary

`get_ppo_ray_runtime_env()` previously returned only rLLM defaults plus host-forwarded env vars. When the same training script was launched via `ray job submit --runtime-env-json=...`, Ray's own merge inside `ray.init()` would raise on any key that appeared in both dicts unless the user exported `RAY_OVERRIDE_JOB_RUNTIME_ENV=1`. This made the obvious workflow ("set `NCCL_DEBUG=INFO` for one cluster run") fail in confusing ways. This PR pins down a single precedence contract — rLLM defaults < host-forwarded < job-submit — and pre-empts the collision by popping any key the job config sets from the dict we return.

## Type of change

- [ ] Feature
- [x] Fix
- [x] Docs
- [ ] Refactor
- [ ] Example / Project
- [ ] Infra / CI

## What changed

- `get_ppo_ray_runtime_env()` now reads `RAY_JOB_CONFIG_JSON_ENV_VAR` and pops every `env_vars` key the job config sets, so Ray's `ray.init()` merge sees no conflicts and the job-submit value wins.
- `working_dir=None` is only added when the job config does not set one (avoids the same merge conflict on `working_dir`).
- Malformed job-config JSON is tolerated — falls back to the previous behavior instead of raising.
- Tests cover: no job config, key-pop on overlap, `working_dir` skip, and invalid JSON.
- New docs page `docs/guides/ray-runtime-env.mdx` (added to the `Guides` group in `docs.json`) writes down the precedence contract and the override paths.

## Validation

- [x] Targeted tests: `pytest tests/trainer/verl/test_ray_runtime_env.py`
- [ ] Not run (reason below)

Validation details:
- Existing forwarding tests still pass; four new tests added for the job-submit merge cases.
- `mint broken-links` clean for the new docs page.

- None for users who do not launch via `ray job submit`. For users who do, a `--runtime-env-json` `env_vars` entry now silently overrides the matching rLLM default instead of failing the job — this is the intended behavior and the previously-needed `RAY_OVERRIDE_JOB_RUNTIME_ENV=1` workaround is no longer required.

## Docs / examples

- [ ] Not needed
- [x] Updated docs
- [ ] Updated examples
- [ ] Follow-up docs needed

Includes new page `docs/guides/ray-runtime-env.mdx` covering the precedence contract, default env vars, forward prefixes, `RLLM_EXCLUDE`, and the `ray job submit --runtime-env-json` override path. Wired into the `Guides` group in `docs/docs.json`.

## Related issues / PRs

- Fixes #
- Related to #
- Stacked on / depends on #

## Screenshots / logs

n/a

### DIFF
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -89,7 +89,8 @@
             "group": "Guides",
             "pages": [
               "guides/customizing-training",
-              "guides/distributed-training"
+              "guides/distributed-training",
+              "guides/ray-runtime-env"
             ]
           },
           {
--- a/docs/guides/ray-runtime-env.mdx
+++ b/docs/guides/ray-runtime-env.mdx
@@ -1,0 +1,115 @@
+---
+title: Ray runtime environment
+description: How rLLM builds the Ray runtime_env that controls environment variables on training workers, and how to override it
+---
+
+When rLLM launches verl training, it calls `ray.init()` with a `runtime_env` that controls which environment variables every Ray worker sees. This page describes what `get_ppo_ray_runtime_env()` puts in that dict, the precedence order between rLLM defaults, your shell, and `ray job submit --runtime-env-json`, and the knobs you have for overriding each layer.
+
+## Precedence
+
+`get_ppo_ray_runtime_env()` composes the runtime env from three layers. Higher layers win on key conflicts.
+
+| Priority | Source | Where it comes from |
+| -------- | ------ | ------------------- |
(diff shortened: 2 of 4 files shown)


listar2000:
Also added the missing docs on RAY runtime envs.

PR #539, opened 2026-05-02
refactor(verl): detach VerlBackend from RayPPOTrainer

kylemontgomery1:
## Summary

`VerlBackend` no longer inherits from `RayPPOTrainer`. It composes verl primitives directly via `_init_colocated_workers` plus helpers extracted into a new `rllm/experimental/verl/utils.py` (`build_wg_kwargs`, `save_checkpoint`, `load_checkpoint`, `balance_batch`, `start_profiling`, `stop_profiling`, `create_dataloaders`).

`VerlTaskRunner` (renamed from `WorkflowTaskRunner`) drops `add_critic_worker` and `load_reward_manager` — reward-fn loading is no longer plumbed since the rLLM path computes rewards directly inside the workflow.

## Type of change

- [ ] Feature
- [ ] Fix
- [ ] Docs
- [x] Refactor
- [ ] Example / Project
- [ ] Infra / CI

## What changed

- **Detached lifecycle.** `VerlBackend` is now `BackendProtocol` only. The `RayPPOTrainer.init_workers` body is reproduced inside `_init_colocated_workers`, with helpers (`save_checkpoint`, `load_checkpoint`, `balance_batch`, `build_wg_kwargs`, `create_dataloaders`, `start_profiling`/`stop_profiling`) lifted into `rllm/experimental/verl/utils.py` so they're testable without a `RayPPOTrainer` instance.
- **`VerlTaskRunner` rename.** `WorkflowTaskRunner` is now `VerlTaskRunner`, and no longer adds a critic worker or loads a reward manager.
- **Opinionated choices baked into the rLLM-on-verl path** (now documented in `docs/backends/verl.mdx → How rLLM uses verl`):
  - `use_critic` is forced `False` — no critic worker is registered or initialised; value-based estimators (GAE, REMAX) cannot run as-is.
  - `reward.reward_model.enable=True` is **rejected at startup** (assertion in `VerlBackend.validate_config`). Compute rewards inside the workflow.
  - `algorithm.use_kl_in_reward=True` is rejected at startup. KL-in-loss (`actor.use_kl_loss=True`) is unaffected and still runs inside verl's actor worker.
  - `actor_rollout_ref.rollout.mode` must be `async`.
  - `trainer.use_legacy_worker_impl` is forced to `disable`.
[...]

PR #538, opened 2026-05-02
refactor(verl): pass server_manager directly into VerlEngine

kylemontgomery1:
`VerlEngine.__init__` previously took an `AgentLoopManager` and reconstructed an `AsyncLLMServerManager` internally. Move that construction up to the caller (`VerlBackend.init_rollout_engine`) and have `VerlEngine` accept the `AsyncLLMServerManager` directly. This decouples `VerlEngine` from the `AgentLoopManager` plumbing and lets future callers (e.g. async/separated mode) hand in a custom server manager.

Also swap the two `print` calls for `logger.info` and add a `logger` to the module.

## Type of change

- [ ] Feature
- [ ] Fix
- [ ] Docs
- [x] Refactor
- [ ] Example / Project
- [ ] Infra / CI

## What changed

- 
- 
- 

## Validation

- [x] `pre-commit run --all-files`
- [x] Targeted tests: `pytest ...`
- [x] Manual validation performed
- [ ] Not run (reason below)

Validation details:
- 

## Breaking changes / migration notes

- None

## Docs / examples

- [x] Not needed
- [ ] Updated docs
- [ ] Updated examples
- [ ] Follow-up docs needed

## Related issues / PRs

- Fixes #
- Related to #
- Stacked on / depends on #

PR #537, opened 2026-05-02
refactor(verl): drop verl-native advantage path and deprecate use_rllm

kylemontgomery1:
Drop the verl-native advantage path on the experimental unified trainer; every backend now routes through the rLLM-native estimator hook. Take the opportunity to (a) make the estimator signature uniform across estimators (previously GRPO-special), (b) emit a `DeprecationWarning` for `algorithm.use_rllm`, and (c) rewrite the affected docs and add a worked example for porting a Verl-only estimator (OPO).

## Summary

`compute_advantage_verl` and the `algorithm.use_rllm` toggle are gone. Advantages always flow through `collect_reward_and_advantage_from_trajectory_groups` + `update_dataproto_with_advantages`. The estimator hook signature is now `(rewards, algorithm_config, **kwargs)` for every built-in, and the orchestrator injects `traj_groups` via `**kwargs` so length-aware estimators (OPO etc.) can opt in. `algorithm.use_rllm` raises a `DeprecationWarning` from both `AlgorithmConfig.__post_init__` and `from_config`. Docs are rewritten to drop stale `use_rllm` instructions, explain the `list[np.ndarray]` data contract, and walk through a real OPO port from Verl.

## Type of change

- [ ] Feature
- [ ] Fix
- [x] Docs
- [x] Refactor
- [ ] Example / Project
- [ ] Infra / CI

## What changed

- **`verl_backend.compute_advantages`**: drop the `use_rllm` branch and `compute_advantage_verl`. Always call the rLLM-native path.
- **`AlgorithmConfig`**: `use_rllm` is now `bool | None = None`; both `__post_init__` and `from_config` emit a `DeprecationWarning` when it's set.
- **`collect_reward_and_advantage_from_trajectory_groups`**: drop `_prepare_adv_estimator_input`. Pass `(rewards, algorithm_config, traj_groups=...)` to estimators so each pulls what it needs from `algorithm_config`.
- **Built-in estimators (GRPO / REINFORCE / REINFORCE++ baseline / RLOO)**: unified signature `(rewards, algorithm_config, **kwargs)`. GRPO reads `algorithm_config.norm_adv_by_std_in_grpo`; the others ignore the config arg.
- **`register_rllm_adv_estimator`**: docstring documents the canonical signature, the shape of `rewards`, and how to read `traj_groups` from `**kwargs`.
- **Docs**:
[...]

PR #540, opened 2026-05-02
feat(verl): rllm/verl config sync with shared-keys precedence

kylemontgomery1:
Adds `sync_config` to `rllm/experimental/verl/utils.py`. Keeps verl-native and rllm-namespaced config in agreement via a shared-keys table. Precedence per key: rllm CLI explicit > verl CLI explicit > rllm yaml default > verl yaml default. Hydra CLI overrides are captured in the launcher and forwarded into the Ray actor through a `hydra_overrides` kwarg (Hydra context isn't available across processes).

Special handling for KL and clip_ratio is needed.

The verl backend yaml is correspondingly stripped of `???` placeholders and `oc.select` resolvers — the runtime sync handles propagation now.

## Summary

- New `sync_config` runs at trainer startup and bidirectionally mirrors a fixed table of shared keys between the verl-native namespace and the `rllm.*` namespace.
- Precedence per shared key: **rllm CLI explicit > verl CLI explicit > rllm yaml default > verl yaml default**.
- Hydra CLI overrides (`HydraConfig.get().overrides.task`) are captured in the launcher and forwarded as a `hydra_overrides` kwarg into the Ray actor, since `HydraConfig` is not available across processes.
- Special handling for KL (`actor.use_kl_loss = (kl_beta > 0)` when not explicitly set) and the `clip_ratio` family (`rllm.eps_clip ↔ actor.clip_ratio` symmetric, with `clip_ratio_low` priority).
- The verl backend yaml is correspondingly stripped of `???` placeholders and `oc.select` resolvers.
- Adds `tests/unified_trainer/test_config_sync.py` covering the precedence sweep on `adv_estimator`, the KL derivation, and the clip-ratio family.

## Type of change

- [ ] Feature
- [ ] Fix
- [ ] Docs
- [x] Refactor
- [ ] Example / Project
- [ ] Infra / CI

## What changed

- `rllm/experimental/verl/utils.py` — new `_explicit_override_keys`, `_SHARED_KEYS` table, and `sync_config` function.
- `rllm/experimental/verl/verl_launcher.py` — captures `hydra_overrides` in the Hydra-decorated process and forwards them into `VerlTaskRunner.run`, which calls `sync_config` after `OmegaConf.resolve`.
- `rllm/experimental/config/rllm/backend/verl.yaml` — drops the `???` placeholders and `oc.select` resolvers; only the genuine verl-side defaults that must differ from `ppo_trainer.yaml` remain.
- `tests/unified_trainer/test_config_sync.py` — 14 tests for the precedence sweep, KL derivation, and clip-ratio family.
- `docs/experimental/configuration.mdx` — rewrites the "Verl backend" and "Config forwarding mechanism" sections to describe the new bidirectional sync, the precedence ladder, the shared-keys table, and the derived rules (use_kl_loss, clip_ratio).
- `docs/backends/verl.mdx` — adds one bullet to the "How rLLM uses verl" section pointing readers at the new precedence rules.

## ⚠️ Reviewer warning — yaml-overlay regression

`_explicit_override_keys` only captures **CLI** overrides (`HydraConfig.get().overrides.task`), not values introduced by the Hydra `defaults:` chain or `+experiment=...` config-group patterns. For shared keys whose rllm-side default in `base.yaml` is non-`null`, this means a yaml-overlay value on the verl-native side gets silently overwritten:

```yaml
# my_experiment.yaml (loaded via +experiment=my_experiment, no CLI flags)
actor_rollout_ref:
  rollout:
    n: 16
```

Pre-PR (with `${oc.select:actor_rollout_ref.rollout.n, 8}` interpolation), the rllm-side picked up the overlay value 16. Post-PR, `_explicit_override_keys` returns `{"experiment"}` — neither shared-key path is in the explicit set — so `sync_config` falls into the default branch:

```python
value = OmegaConf.select(config, "rllm.rollout.n")  # 8 (rllm base.yaml default)
if value is None: continue
OmegaConf.update(config, "actor_rollout_ref.rollout.n", 8, merge=False)  # clobbers 16 → 8
```

This affects every shared key with a non-`null` rllm `base.yaml` default: `adv_estimator`, `norm_adv_by_std_in_grpo`, `eps_clip`, `kl_beta` (and the derived `use_kl_loss`), `n`, `n_val`, all `trainer.*` cadence/logging knobs, etc.

**Cookbook scripts (`cookbooks/*/train_verl.sh`) are not affected** — they all override on the CLI, which is correctly captured. The regression is specific to users who structure experiments via Hydra config groups instead of CLI flags.

**Suggested fix** (suggested by Claude Code): set the shared-key entries in `rllm/experimental/config/rllm/base.yaml` to `null`, push the actual fallbacks into the config dataclasses' `__post_init__` / `from_config`, and rely on the existing `value is None → continue` branch to leave the verl-native value alone when no one is on the CLI. This matches the docstring's stated intent ("`None` rllm values are treated as 'no rllm default'") for both CLI and yaml-overlay paths. Add a regression test where `_explicit_override_keys` returns the empty set but the verl-native value differs from the rllm default — current code would clobber, a fixed implementation would preserve.

The PR author should decide whether to address this in this PR or punt it as a follow-up; the regression is silent so I wanted it visible in the PR record either way.

## Validation

- [x] `pre-commit run --files rllm/experimental/verl/utils.py docs/...` (ruff + ruff-format pass)
- [x] Targeted tests: `./venv/bin/python -m pytest tests/unified_trainer/test_config_sync.py -q` → 14 passed
- [x] Docs: `mint validate` (build passes) and `mint broken-links` (no broken links) inside `docs/`
- [ ] Not run (reason below)

Validation details:
- The 14 unit tests cover the precedence sweep on `adv_estimator`, KL derivation (kl_beta zero/positive, explicit override skipping derivation), and the `clip_ratio` family (default propagation, rllm→verl, verl→rllm, low-takes-precedence-over-symmetric, asymmetric clip).
- I have not added the regression test described in the warning section above.
- Unit tests do not exercise an end-to-end Ray run; recommended manual smoke before tagging release: a single training step driven entirely from the rllm-namespace side (`rllm.algorithm.adv_estimator=rloo rllm.rollout.n=4 rllm.algorithm.kl_beta=0.01`) to confirm the sync produces the same trainer behaviour as the equivalent verl-native CLI.

- `actor.use_kl_loss` is now derived from `kl_beta > 0` whenever the user does not explicitly set it on the CLI. Users who previously set `actor.kl_loss_coef` (or `rllm.algorithm.kl_beta`) without also setting `use_kl_loss=True` will now get KL-in-loss enabled (was silently disabled before). Almost certainly the user's intent, but call it out.
- See "⚠️ Reviewer warning — yaml-overlay regression" above for a behavioural difference that affects users overriding shared keys via Hydra config groups instead of CLI flags.

## Docs / examples

- [ ] Not needed
- [x] Updated docs
- [ ] Updated examples
- [ ] Follow-up docs needed

## Related issues / PRs

- Stacked on / depends on #538, #539
- Related to #537, #546

## Screenshots / logs
Screenshot 2026-05-05 at 6 12 39 PM


listar2000:
This is currently conflicted due to some removal of GAE-related configs in #546 -- will fix this conflict later.

listar2000:
Taking a look rn. Will also work on solving the conflict.

listar2000:
@kylemontgomery1 overall this looks good! Only concern/regression is how this works with `yaml-override` (most of our cases is CLI/script-based override). Please see my edit to the PR above.

Other than that I've fixed merge conflict and updated the docs a bit.

If you think this is not a huge concern, feel free to merge it!

kylemontgomery1:
@listar2000 so the issue is basically that verl-style shared keys changed by yaml config files can be mistaken for verl defaults and overwritten by rLLM defaults?

listar2000:
> @listar2000 so the issue is basically that verl-style shared keys changed by yaml config files can be mistaken for verl defaults and overwritten by rLLM defaults?

@kylemontgomery1 yeah basically as you said (I'm also running some local experiment rn for this) -- this might affect ppl who might be writing their own `hydra` yaml file instead of using our provided `unified.yaml`.

listar2000:
What do you think about the suggested fix of *setting shared-keys within `base.yaml` to have `null` values*?

So that we will set those defaults in the actual Python config files -- also echoing our philosophy that "`null` rllm values are treated as 'no rllm default'"

PR #543, opened 2026-05-02
feat(verl): support R2 and R3 MoE router replay

kylemontgomery1:
Widen rllm.algorithm.router_replay from bool to Literal["disabled", "R2", "R3"]. R2 records routing during the proximal compute_log_prob forward pass (training-only); R3 records at rollout and ships the data through the batch.                                                                                                 
   
For R3, encode routed_experts as [shape_header_json, base64_blob] per step on the rollout side. Each step's routing covers (step.prompt + step.action). In transform.py, since segments are cumulative by construction, the last step's routing covers the entire segment's full sequence — decode at emit time and place at [max_prompt_length - len(prompt) : max_prompt_length + len(response)] in the padded (bs, prompt_len + resp_len, num_layers, topk) tensor, mirroring verl's agent_loop start_pos:end_pos layout.
                                                                                                            
For R2, propagate routed_experts from compute_log_prob's output into the batch (matching verl's native pattern in ray_trainer._compute_old_log_prob). Without this, the proximal forward's recorded routing is dropped before the actor update reads it. Also includes entropys in the old-log-prob tensordict and pops it after the entropy metric, mirroring verl native.                                                            
 
sync_config mirrors the mode to actor.router_replay.mode and sets rollout.enable_rollout_routing_replay=True for R3. VerlBackend validation requires actor.strategy == "megatron". TinkerBackend rejects router_replay != "disabled" since tinker doesn't support it.

## Type of change

- [ ] Feature
- [ ] Fix
- [ ] Docs
- [ ] Refactor
- [ ] Example / Project
- [ ] Infra / CI

## What changed

- 
- 
- 

## Validation

- [ ] Not run (reason below)

Validation details:
- 

- None

## Docs / examples

- [ ] Not needed
- [ ] Updated docs
- [ ] Updated examples
- [ ] Follow-up docs needed

## Related issues / PRs

- Fixes #
- Related to #
- Stacked on / depends on #


listar2000:
Reviewing this PR rn. But just for some context @kylemontgomery1 -- for supporting router replay in Verl, can I assume that in rLLM side we just need to pass around / adapt some configs and can delegate the rest to Verl? Or is it more like the fully async case that we need to write our own impl?

rllm/experimental/verl/utils.py
```diff
@@ -195,6 +196,11 @@ def sync_clip_ratio() -> None:
         if kl_beta is None:
             kl_beta = 0.0
         OmegaConf.update(config, "actor_rollout_ref.actor.use_kl_loss", kl_beta > 0, merge=False)
+
```
listar2000: @kylemontgomery1 Can you take a look at this place? I just merged some conflicts. Want to double check that these router-related configs are still compatible with our recent sync-changes.

listar2000:
This PR LGTM -- other than maybe @kylemontgomery1 you can double-check on my merges back from main.

rllm/experimental/verl/verl_launcher.py
```diff
@@ -34,6 +34,7 @@ def run(self, config, workflow_class: type[Workflow], workflow_args: dict, hydra
         OmegaConf.register_new_resolver("mul", lambda x, y: int(x) * int(y))
         sync_config(config, hydra_overrides=hydra_overrides)
         OmegaConf.resolve(config)
+        sync_config(config, hydra_overrides=hydra_overrides)
```
listar2000, line 37: Are we syncing twice here?

listar2000:
Double-checked myself and ready to merge.

PR #544, opened 2026-05-02
feat(verl): emit offpolicy/* metrics for both bypass and non-bypass paths

kylemontgomery1:
Whenever rollout_log_probs is in the batch, emit a block of off-policy diagnostics computed by verl's compute_offpolicy_metrics (KL, log-PPL difference, ppl ratio, pearson correlation, chi-squared) under the offpolicy/* namespace. Runs for both bypass mode (where π_old = π_rollout, so the metrics reduce to identity diagnostics) and the standard non-bypass path. Move calculate_debug_metrics_compat to the same top-level position so both paths get the legacy training/rollout_probs_diff_* debug metrics too.

## Type of change

- [x] Feature
- [ ] Fix
- [ ] Docs
- [ ] Refactor
- [ ] Example / Project
- [ ] Infra / CI

## What changed

- 
- 
- 

## Validation

- [ ] Not run (reason below)

Validation details:
- 

- None

## Docs / examples

- [x] Not needed
- [ ] Updated docs
- [ ] Updated examples
- [ ] Follow-up docs needed

## Related issues / PRs

- Fixes #
- Related to #
- Stacked on / depends on #

### DIFF
--- a/rllm/experimental/config/rllm/backend/verl.yaml
+++ b/rllm/experimental/config/rllm/backend/verl.yaml
@@ -1,63 +1,32 @@
 # @package _global_
 
-# Verl Backend Configuration for rLLM
-# This config largely follows from the `verl.trainer.config.ppo_trainer.yaml`
-# with the additional "overriding" of rLLM-common config with Verl-specific ones (backward compatibility)
+# Verl Backend Configuration for rLLM.
+# Loads verl's ppo_trainer.yaml as defaults; rllm-namespaced values are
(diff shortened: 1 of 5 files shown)


listar2000:
LGTM -- but I imagine that this would be another metrics discrepancy with `tinker` side right? Or shall we also implement similar metrics for the tinker as well (essentially since bypass = True, just identity diagnostics -- but might be repetitive with some existing ones).

PR #548, opened 2026-05-04
fix(verl+geo3k): make Qwen3-VL geo3k cookbook actually train

listar2000:
## Summary

Makes the `cookbooks/geo3k` cookbook actually trainable end-to-end on the **Verl backend** with `Qwen/Qwen3-VL-2B-Instruct`. Branched from `main` after a long iterative debug session against an 8×H100 box. Five distinct issues had to be fixed; details below.

**Tinker backend with geo3k is *not* tested in this PR** — only Verl. The data-layer fix (parquet round-trips images) and the geo3k_flow `<image>` placeholder fix are backend-agnostic and should also benefit the Tinker path, but I haven't run `geo3k_local_tinker.sh` to confirm. Leaving that for a follow-up.

## What was broken

| # | Layer | Symptom | Resolution |
| - | ----- | ------- | ---------- |
| 1 | rLLM dataset | `apply_verl_postprocessing` stripped binary columns to `None`, so `extra_info[\"images\"]` was empty in the verl parquet — the agent flow received no images and the model trained on text-only prompts. | Replace `_strip_binary_columns` with `_wrap_binary_columns_for_parquet` that wraps each `bytes` as `{\"bytes\": ..., \"path\": \"\"}` (verl's canonical image format). Images flow through `extra_info[\"images\"]` for rLLM agent flows and are also valid for verl-native consumers. |
| 2 | geo3k_flow | The dataset's `question` field has a literal `<image>` placeholder at the start (`\"<image>Find x.\"`). The old `_build_vlm_content` appended that text *as-is* alongside an `image_url` block — vLLM's chat template then rendered both the expanded image AND the literal `<image>` token, producing a malformed prompt. | Mirror verl's `rl_dataset._build_messages`: split the user content on `(<image>\|<video>)` markers and interleave `image_url` blocks at marker positions. Strict `<image>`-count check; centralized `_img_to_data_uri` helper handles raw bytes / `{\"bytes\": ...}` dict / data URI / PIL. |
| 3 | verl 0.7.1 | `Conv*d`/SDPA crashed with `CUDNN_STATUS_NOT_INITIALIZED` on this host (4-line stock-torch reproducer; nothing to do with rLLM/verl). | Not committed to rLLM. The fix is now an environment-specific worker setup hook the user can plug in; see the new `RLLM_EXTRA_WORKER_SETUP_HOOK` extension point below. |
| 4 | verl 0.7.1 | `prepare_dynamic_batch` deadlocks on imbalanced micro-batch counts across DP ranks ([verl#5750](https://github.com/volcengine/verl/issues/5750)); inplace `+=` in the qwen3-VL no-image dummy path crashes ([verl#5881](https://github.com/volcengine/verl/pull/5881)); `index_select_tensor_dict` / `chunk_tensordict` lose `_ragged_idx` when sample seq lengths coincide ([verl#6127](https://github.com/volcengine/verl/pull/6127), causing both the dynamic-batching `unbind` failure and the rmpad rotary shape mismatch). | Three idempotent monkey-patches in `rllm/experimental/verl/patch.py`, applied in **every Ray worker process** via `runtime_env.worker_process_setup_hook` — driver-side mutation does not propagate because Ray workers re-import verl from disk. |
| 5 | rLLM Ray runtime env | Driver-only patches were silently a no-op for the actual call sites. | Register `rllm.experimental.verl.patch.apply_all_verl_patches` as `runtime_env.worker_process_setup_hook` in `get_ppo_ray_runtime_env()`. Add `RLLM_` to `FORWARD_PREFIXES` so user extension hooks can be wired via env var. |

## Code changes

- `rllm/data/dataset.py` — `_strip_binary_columns` → `_wrap_binary_columns_for_parquet`. Images now survive the `.parquet` round-trip as `{\"bytes\": <png-bytes>, \"path\": \"\"}` dicts.
- `cookbooks/geo3k/geo3k_flow.py` — proper `<image>`-marker interleaving (mirrors verl), handles all three image shapes, raises on marker/image mismatch.
- `rllm/experimental/verl/patch.py` — three new patches: `patch_verl_qwen3_vl_dummy_inplace` ([verl#5881](https://github.com/volcengine/verl/pull/5881)), `patch_verl_tensordict_jagged_layout` ([verl#6127](https://github.com/volcengine/verl/pull/6127)). Existing `patch_verl_dynamic_batch_sync` ([verl#5750](https://github.com/volcengine/verl/issues/5750)) was already there but driver-only and effectively a no-op; now reachable from the worker hook. Plus `apply_all_verl_patches` umbrella + `RLLM_EXTRA_WORKER_SETUP_HOOK` extension point for environment-specific tweaks (file-path-based loader so user hooks can live in a gitignored `tmp/` dir).
- `rllm/experimental/verl/verl_backend.py` — drop the two no-op driver-side patch calls (`patch_verl_dynamic_batch_sync` + `patch_verl_qwen3_vl_dummy_inplace` only matter inside workers); keep one driver-side `patch_verl_tensordict_jagged_layout()` call because `DataProto.chunk` runs in the driver and ships chunks with baked-in `_ragged_idx`. `patch_vllm_for_sdk` stays driver-side (Ray pickles the actor class, so driver-side mutation is correct there).
- `rllm/trainer/verl/ray_runtime_env.py` — register the worker hook; add `RLLM_` to `FORWARD_PREFIXES`.

## Docs

- `docs/guides/ray-runtime-env.mdx` — new section documenting the worker setup hook (why driver-only mutation doesn't work, and how `apply_all_verl_patches` is wired) and the `RLLM_EXTRA_WORKER_SETUP_HOOK` extension point.

## Tests

- `tests/data/test_arrow_ipc.py` — renamed `TestStripBinaryColumns` → `TestWrapBinaryColumnsForParquet`. Coverage for scalar bytes, list-of-bytes, and `None`.

## Verification

Before, the gateway trace had:
```
user: '<image>Find $PS$.'   ← string content, no image_url block
prompt_token_ids count: 0
```

After:
```
user: [image_url block (data URI), text='Use parallelogram $JKLM$ to find $m \angle KJL$.']
prompt_token_ids count: 328
<|image_pad|> count: 240
literal '<image>' substring? False
user portion: <|im_start|>user\n<|vision_start|><|image_pad|>×240<|vision_end|>Use parallelogram...
```

This matches the canonical Qwen3-VL prompt shape. Training reached `Generating trajectories: 76/256 @ 4-5it/s` cleanly on the same host (with the env-specific cuDNN-disable hook in place via `RLLM_EXTRA_WORKER_SETUP_HOOK`); I stopped early once the prompt-layer fix was confirmed at the trace level.

## Out of scope / left for follow-up

- **Tinker backend with geo3k**: untested. The data-layer + flow-layer fixes are backend-agnostic and should apply, but I haven't run it.
- **Broken host cuDNN**: not an rLLM bug. Real fix is a system-level cuDNN/torch update; the workaround lives in user space via `RLLM_EXTRA_WORKER_SETUP_HOOK`.
- **Top-level `images` for verl-native pipeline**: not added. Doing so requires the `prompt` field (currently a placeholder) to also have `<image>` markers, which is out of scope for the agent-flow fix.

## Test plan

- [x] CI green
- [ ] `bash tmp/geo3k/geo3k_local.sh` (or your equivalent) reaches step ≥1 cleanly with vision tokens in the gateway trace
- [x] `pytest tests/data/test_arrow_ipc.py` passes (verified locally)
- [ ] (follow-up) Tinker backend with geo3k

### DIFF
--- a/cookbooks/geo3k/geo3k_flow.py
+++ b/cookbooks/geo3k/geo3k_flow.py
@@ -9,6 +9,7 @@ from __future__ import annotations
 
 import base64
 import logging
+import re
 
 from openai import AsyncOpenAI
 
@@ -31,7 +32,7 @@ async def geo3k_flow(task: Task, config: AgentConfig) -> Episode:
     """Single-turn VLM geometry solver."""
     client = AsyncOpenAI(base_url=config.base_url, api_key="EMPTY")
     question = task.instruction
-    images = task.metadata.get("images", [])
+    images = task.metadata.get("images") or []
 
     user_content = _build_vlm_content(question, images) if images else question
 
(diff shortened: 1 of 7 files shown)


listar2000:
@jeffreysijuntan since you've refactored the `dataset` class for agent flow earlier, might bother you to take a look at the related fixes. The other `Verl` patches should be all good (they mirror a few post-0.7.1 PRs in the Verl repo).

jeffreysijuntan:
lgtm!

PR #551, opened 2026-05-05
feat(model-gateway): trace store schema v2

jeffreysijuntan:
## Summary

First of three stacked PRs splitting the gateway upgrade. Standalone — no consumer changes; landed alone, the new schema and methods are dead code that the next two PRs activate.

- Denormalises hot filter columns (`run_id`, `model`, `harness`, `latency_ms`, `has_error`, `step_id`) onto the `traces` row so the cross-run trace feed (rllm-console PR) can filter without JSON-scanning every row.
- Adds a `runs` table for per-gateway-run lifecycle.
- Extends `TraceStore` protocol: `query_traces`, `register_run`, `end_run`, `list_runs`, `count_traces`, `facets`.
- Schema-version mismatch on an existing db drops & recreates every table — gateway data is dev-only and a real migration isn't worth it.
- `_get_conn` serialises first-call init under an `asyncio.Lock` so a burst of concurrent `store_trace` callers can't race the schema build (the previous lock-free path could leave racers staring at a half-dropped db when migration triggered).
- Memory store mirrors the new shape so unit tests cover both backends identically.

Existing `store_trace(trace_id, session_id, data)` callers keep working — `run_id` defaults to `""`, all new methods are additive.

## Stack

- **[PR-1A] this** — schema v2 + lock fix
- [PR-1B](https://github.com/rllm-org/rllm/pull/new/pr-1b-metadata) — `X-RLLM-*` headers + inbound bearer auth (depends on this)
- [PR-1C](https://github.com/rllm-org/rllm/pull/new/pr-1c-upstream) — upstream-proxy mode + run lifecycle (depends on B)

## Test plan

- [x] `python -m pytest tests/unit/` — 220 passed
- [x] `python -m pytest tests/unit/test_store.py` — 74 passed (new tests for schema v2, denormalised columns, junction back-compat, concurrent writers)
- [ ] One known flaky case: `test_two_stores_share_one_db` can fail ~5% under suite-wide ordering when two `SqliteTraceStore` instances race the WAL pragma. Passes reliably in isolation. Tracking as a follow-up — fix is a small retry on `journal_mode=WAL` or a process-level init lock.

### DIFF
--- a/rllm-model-gateway/src/rllm_model_gateway/store/base.py
+++ b/rllm-model-gateway/src/rllm_model_gateway/store/base.py
@@ -9,10 +9,20 @@ class TraceStore(Protocol):
     Implementations must be async.  The interface uses plain dicts so that
     backends are free to serialise however they like (JSON columns, DynamoDB
     items, etc.).
+
+    Traces are stamped with ``run_id`` so multiple gateway instances can
+    share a single store without colliding session ids. ``run_id=""`` is
+    the unstamped bucket for legacy callers.
     """
 
-    async def store_trace(self, trace_id: str, session_id: str, data: dict[str, Any]) -> None:
-        """Store a single trace."""
+    async def store_trace(
+        self,
+        trace_id: str,
+        session_id: str,
+        data: dict[str, Any],
+        run_id: str = "",
+    ) -> None:
+        """Store a single trace, optionally tagged with ``run_id``."""
         ...
 
     async def get_trace(self, trace_id: str) -> dict[str, Any] | None:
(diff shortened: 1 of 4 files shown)


luyuzhe111:
@jeffreysijuntan look good overall. what's the PR that tackles the flaky test tho?

jeffreysijuntan:
> @jeffreysijuntan look good overall. what's the PR that tackles the flaky test tho?

It is in a new uncommit PR, I will push it to here instead

listar2000:
What's the current blocker for this PR to get merged (the CI failure?)

jeffreysijuntan:
> What's the current blocker for this PR to get merged (the CI failure?)

Before changing the gateway, I am working on some PR to unify the eval path with the AgentFlowEngine, so eval will run directly through that in the future.

PR #554, opened 2026-05-05
examples(agentcore_math): enable bypass mode and make verl example fit on 8xA100 (40GB)

luyuzhe111:
## Summary

Update the verl agentcore_math example to (1) match the tinker sibling's rollout-correction posture via bypass mode, and (2) broaden hardware support so the script now runs on 8 x 40GB+ GPUs (verified on 8x A100 40GB). (3) fix a couple issues that come with higher TP degree.

### `examples/agentcore_math/train_agentcore_math_verl.sh`

- **Enable bypass mode** (`rllm.algorithm.rollout_correction.bypass_mode=true`) so π_old = π_rollout, skipping the actor's log-prob recomputation. This mirrors what the tinker sibling script effectively does by construction (tinker's `training_logprobs` come from the sampler pass). Trade-off: drops `actor/entropy` and `training/rollout_probs_diff_*` metrics, since both are computed only on the non-bypass path in `verl_backend.py`. All loss/advantage/reward/timing metrics are preserved.
- **Drop deprecated `rllm.algorithm.use_rllm=true`.** Now a no-op with DeprecationWarning since #537.
- **Work around vllm-project/vllm#34932** via `+actor_rollout_ref.rollout.engine_kwargs.vllm.tokenizer_mode=slow`. Avoids the `Already borrowed` race in `Hermes2ProToolParser.__init__` under concurrent `/v1/chat/completions` load, which caused ~2% of rollout requests to 400 and occasionally produced empty trajectories. Upstream fix is vLLM PR #38168, first released in vllm 0.18.1; inline comment notes to remove this override once rllm's `vllm==0.17.0` pin bumps past 0.18.1. This issue happens when TP>1.
- **Set `rollout.max_model_len=16384`** to match `max_prompt_length + max_response_length`.

### `rllm/experimental/common/visualization.py`

- Guard `response_ids[-1]` access in `visualize_trajectory_last_steps`. Previously crashed the trainer when a trajectory had an empty response. This is orthogonal to the vLLM tokenizer race — agent timeouts or tool errors can produce the same shape. Now prints `<empty response>` and continues.

## Tests

- [x] Run on 8x A100 40GB — completed `total_epochs=1`. Training was stable and converges fast. 
- [x] Verified bypass mode is in effect by absence of `actor/entropy` and `training/rollout_probs_diff_*` in logs (both gated on the non-bypass branch in `verl_backend.py`).
- [x] Zero `Already borrowed` / 400 Bad Request errors under the slow-tokenizer override.

Screenshot 2026-05-05 at 11 54 35 AM

### DIFF
--- a/examples/agentcore_math/train_agentcore_math_verl.sh
+++ b/examples/agentcore_math/train_agentcore_math_verl.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 # Requires megatron deps: bash scripts/install_megatron.sh <cu128|cu129|cu130|...>
+# Tested on 8x A100 40GB. Parallelism (TP=8, SP=true) and gpu_memory_utilization
+# are tuned for this configuration; larger GPUs likely support lower TP.
 set -eux
 
 # Load environment variables (AGENTCORE_AGENT_ARN, AGENTCORE_S3_BUCKET)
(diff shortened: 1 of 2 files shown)


listar2000:
LGTM!

PR #568, opened 2026-05-07
fix(verl): release GPU memory on Verl backend interrupts

listar2000:
## Summary

A crashed or Ctrl-C'd verl run leaves vLLM rollout replicas awake on every GPU, holding tens of GB each, plus orphaned `AsyncvLLMServer` Ray actors with their mp-spawn vLLM worker subprocesses. Root cause: `VerlBackend.shutdown()` was a placeholder, so the launcher's `try/finally` calling `trainer.shutdown()` did nothing for verl-specific resources, and Ray was never explicitly torn down. This PR adds three small cleanup hooks so an interrupted run actually releases its GPU memory.

## Type of change

- [x] Fix

## What changed

- `rllm/experimental/verl/verl_backend.py` — `VerlBackend.shutdown()` now calls `checkpoint_manager.sleep_replicas()` so GPU memory is dropped on every exit path, not just clean training completion. Errors are logged and swallowed so shutdown can't raise.
- `rllm/experimental/unified_trainer.py` — `UnifiedTrainer.fit_async` wraps the training loop in `try/finally` so `backend.on_train_end` runs after a mid-loop crash. The `val_only` and pre-training-validation paths intentionally bypass `on_train_end` to match prior behavior (avoids surprising tinker checkpoint saves on val-only or validation-error paths).
- `rllm/experimental/verl/verl_launcher.py` — `VerlTrainerLauncher.train` now tracks whether it created the Ray cluster and calls `ray.shutdown()` in a `finally` block if so, reaping `AsyncvLLMServer` actors and their vLLM mp child workers together. If Ray was already running on entry, it's left alone.

## Validation

- [x] `pre-commit run --all-files` (ruff + ruff-format)
- [ ] Targeted tests: not run

Validation details:
- Verified the chain end-to-end on a 4×H100 host with Qwen3-0.6B / math500: full val (math500 mean 0.597, accuracy 59.8%, 86s) followed by step 1 (reward 0.43, grad_norm 0.28, throughput 2290 tok/s, update_weights 1.4s) — confirms the changes don't regress the happy path. Crash-path coverage relies on `sleep_replicas` being idempotent (it is) and `ray.shutdown()` being a no-op when we don't own the cluster.

- `UnifiedTrainer.fit_async` now calls `backend.on_train_end` from a `finally` block when training raises mid-loop. For `VerlBackend` this is a no-op (default protocol method). `TinkerBackend.on_train_end` saves a final checkpoint — that now runs on crash too. We kept the `val_only` and pre-training-validation paths the same as before (they still skip `on_train_end`) to avoid the more surprising behavior change of saving a checkpoint after a `val_only` run or after validation itself raised.

## Docs / examples

- [x] Not needed

## Related issues / PRs

- Stacked-on-top draft (ZMQ-jobid IPC backport, separate design discussion): coming as a follow-up PR.

### DIFF
--- a/rllm/experimental/unified_trainer.py
+++ b/rllm/experimental/unified_trainer.py
@@ -320,10 +320,13 @@ class UnifiedTrainer:
         # we start from step (1 + original start batch index)
         trainer_state.global_step += 1
 
-        # Run the training loop
-        await self._fit_async(trainer_state)
-
-        await self.backend.on_train_end(trainer_state)
+        try:
+            await self._fit_async(trainer_state)
+        finally:
+            try:
+                await self.backend.on_train_end(trainer_state)
+            except Exception:
+                logger.exception(f"{self.backend.__class__.__name__}.on_train_end() failed during fit_async() cleanup")
 
     async def _fit_async(self, trainer_state: TrainerState) -> None:
         """Dispatch to sync or concurrent training based on config."""
--- a/rllm/experimental/verl/verl_backend.py
+++ b/rllm/experimental/verl/verl_backend.py
@@ -618,8 +618,17 @@ class VerlBackend(BackendProtocol[Iterable, DataProto]):
             _send_actor_update(sub_batch, loss_name)
 
     def shutdown(self) -> None:
-        """Placeholder, just use the BackendProtocol's default shutdown method."""
-        pass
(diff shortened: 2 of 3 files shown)


listar2000:
@kylemontgomery1 this is a quick utility fix that doesn't hurt our other functionalities.

PR #570, opened 2026-05-08
fix: crash on unmatched wildcard in RLLM_EXCLUDE and _str_to_int return type

dashitongzhi:
## Changes

1. **Fix ValueError crash in ** — When  contains a wildcard pattern (e.g. ) that doesn't match any prefix in ,  raises . Wrapped the call in  so unmatched patterns are silently ignored.

2. **Fix  return type annotation** — Changed return type from  to  in  to match the actual return value.


listar2000:
LGTM. Thanks for contributing.

PR #573, opened 2026-05-09
feat(cookbooks): unified rllm.rollout.{train,val} sampling-params pathway

listar2000:
> Stacked on top of #572 (cookbook namespace fixes). Rebase onto `main` once that lands. Follow-up #578 wires the same sampling-params pathway through `rllm eval`.

## Summary

Centralizes rollout sampling parameters at the rllm level. AgentFlows now read `temperature` / `top_p` / `max_tokens` from `config.sampling_params` — spread straight into their LLM calls — instead of hardcoding them; both training backends source from the same `rllm.rollout.{train,val}` block in `base.yaml`.

## Why

Previously each cookbook flow hardcoded sampling values (`temperature=0.6`, `max_tokens=2048`, …) inside `client.chat.completions.create(…)`. Two consequences:

1. No way for a user to change them without editing the cookbook source.
2. The two backends had their own parallel config: Tinker had `sampling.{train,val}` in `backend/tinker.yaml`; Verl picked up `actor_rollout_ref.rollout.{temperature,top_p,top_k}` from upstream verl yaml. Drift between them was easy.

## What changed

**Plumbing (rllm core)**
- `AgentConfig` (`rllm/types.py`) gains `sampling_params: dict`.
- `AgentFlowEngine.__init__` accepts `train_sampling_params` / `val_sampling_params`, keys them by `is_validation`, and forwards the right dict into each per-rollout `AgentConfig`.
- `UnifiedTrainer` passes `rllm_config.rollout.{train,val}` (⊕ `max_tokens = data.max_response_length`) into the engine.

**Config**
- `base.yaml` adds `rollout.{train,val}.{temperature,top_p,top_k,max_tokens}` (defaults `1.0/1.0/-1/2048`).
- `backend/tinker.yaml` drops the legacy `sampling.{train,val}` block; `TinkerBackend.init_rollout_engine` now reads `full_config.rllm.rollout` (TinkerEngine already destructures `.get("train", {})` / `.get("val", {})` so this slots in).
- `verl/utils.py:_SHARED_KEYS` gains entries mirroring `rllm.rollout.{train,val}.{temperature,top_p,top_k}` ↔ `actor_rollout_ref.rollout.{...}` / `actor_rollout_ref.rollout.val_kwargs.{...}`. Verl-native CLI overrides on these keys keep working but warn deprecated (existing pattern).
- New `sync_do_sample_temperature()` collapses temperature to `0.0` on both sides whenever `do_sample` is False, matching `verl_engine.py`'s engine-init behavior so the rllm-side `sampling_params` reflects the actual sampling rather than a misleading non-zero value.

**Cookbook flows**
- Every flow that previously did `temperature=config.sampling_params.get("temperature", <fallback>)` now builds `sampling = {k: v for k, v in config.sampling_params.items() if k != "top_k"}` and spreads `**sampling` into the LLM call. The `.get(...)` fallback was dead code (the engine always populates `sampling_params`), and `top_k` is dropped because OpenAI-compatible `chat.completions` / langchain `ChatOpenAI` don't accept it.
- The per-cookbook sampling defaults that used to live in those `.get(...)` fallbacks (e.g. `temperature=0.6` for `math` / `deepcoder` / `finqa` / `geo3k`) are now set explicitly in each `train_tinker.sh` — `rllm.rollout.{train,val}.temperature=0.6`. The Verl scripts already set `actor_rollout_ref.rollout.{temperature,val_kwargs.temperature}`, which `sync_config` mirrors into `rllm.rollout.{train,val}.temperature`. `max_tokens` continues to come from `data.max_response_length` (merged in by `UnifiedTrainer`), so no script change is needed for it.
- Frameworks whose SDKs abstract sampling away (`openai_agents`, `smolagents`, `strands`) are left as-is — exposing their knobs requires per-framework wiring out of scope here (only `langgraph`, whose `ChatOpenAI` ctor takes the spread directly, is updated).
- `solver_judge_flow` call-site fix: helper signatures were already updated to take `config: AgentConfig` but the top-level flow still passed `config.model`. Fixed.

## Usage

User can now override per-split via the rllm-native paths:

```bash
python cookbooks/solver_judge_flow/train.py \
    rllm/backend=verl \
    rllm.rollout.train.temperature=1.0 \
    rllm.rollout.val.temperature=0.7 \
    +rllm.rollout.train.max_tokens=8192 \
    +rllm.rollout.val.max_tokens=8192 \
    ...
```

## Test plan

- [x] `pytest cookbooks/<each>/test.py`: green for the cookbooks whose deps are installed locally (finqa needs `asteval`, installed in CI)
- [x] `pytest tests/engine/test_agentflow_engine.py` and the `tests/eval/*` suite: pass
- [x] Hydra compose for both backends: `rllm.rollout.{train,val}` block present, no top-level `sampling` on tinker
- [x] `sync_config` round-trip on verl: rllm overrides propagate to verl paths and vice versa (with deprecation warning)
- [x] `do_sample=False` collapses temperature to `0.0` on both sides
- [x] 4×H100 verl demo on solver_judge_flow with `+rllm.rollout.train.max_tokens=8192`: val + 2 train steps land cleanly. Mean solver response length climbs **1289 → 4809 tokens**, max **1606 → 6288** — directly attributable to the override flowing end-to-end. Truncation ratio **50% → 0%**. Throughput 1067 tok/s, 55s/step.

## Out of scope / follow-ups

- **`rllm eval` plumbing**: forwarding `sampling_params` (and `--temperature` / `--top-p` / `--max-tokens` flags) through `run_dataset` into `AgentFlowEngine` so eval matches training — split out into #578 (stacked on this PR).
- **`agent_frameworks` cookbook**: `openai_agents.py` / `smolagents.py` / `strands.py` build their model wrappers with framework defaults and ignore `config.sampling_params` — exposing their sampling knobs requires per-framework wiring. TODO.
- **Verl `train_*.sh` scripts** still use the verl-native `actor_rollout_ref.rollout.{temperature,val_kwargs.temperature}` form (works via `sync_config`, just emits a deprecation warning). Could be converted to the rllm-native `rllm.rollout.{train,val}.temperature` form for consistency. TODO.
- **`solver_judge_flow` `max_tokens`** now derives from `data.max_response_length` (`=1024` in the train scripts) rather than the `2048` literal it briefly carried; bump `data.max_response_length` if a larger judge/solver budget is wanted.
- The earlier demo run's 0% solver reward was root-caused as a separate agent-flow engine bug (action dropped on trace merge), fixed separately on `main`.


listar2000:
Hi @kylemontgomery1 hope to get you review on this PR as it involves some updates to the configs.

Namely in order for `agent_flow` to have their temperature controlled via config, we introduced `rllm.rollout` config group (which I believe should be something elevated up and become backend-agnostic anyway) that contains sampling params like `temperature`, `top_p`, `top_k`, etc.

The only tricky and confusing part (to be modified in the future) is that the `max_tokens/max_response_length` param now also exists under the `data` config group. So for now:

- The `workflow` (for both Verl and Tinker) uses the `data.max_response_length` for controlling rollout engines.
- The `agentflow` uses the `rllm.rollout.{train, val}.max_tokens` for chat completions.

I make sure these two parameters are synced. Maybe in the future we can further refactor this.

rllm/experimental/config/rllm/base.yaml
```diff
@@ -29,10 +29,26 @@ workflow:
...
+# parameters; ``rllm/experimental/verl/utils.py:sync_config`` mirrors
+# ``rollout.{train,val}.*`` into ``actor_rollout_ref.rollout`` and
+# ``actor_rollout_ref.rollout.val_kwargs`` respectively.
+# The AgentFlowEngine forwards ``rollout.train`` / ``rollout.val`` into
+# ``AgentConfig.sampling_params``, so AgentFlows read user-set sampling
+# values via ``config.sampling_params.get(...)``.
 rollout:
   n: 8
   n_val: 1  # number of validation samples per prompt
+  train:
```
kylemontgomery1, line 43: Can we make this more descriptive, e.g,. either rollout.sampling.{train,val} or rollout.{train,val}_sampling?

rllm/experimental/config/rllm/base.yaml
```diff
@@ -29,10 +29,26 @@ workflow:
...
+    temperature: 1.0
+    top_p: 1.0
+    top_k: -1
+    max_tokens: 2048
```
kylemontgomery1, line 47: This is the same as data.max_response_length?

rllm/experimental/verl/utils.py
```diff
@@ -54,6 +54,12 @@ def _explicit_override_keys(hydra_overrides: list[str] | None = None) -> set[str
     ("actor_rollout_ref.actor.router_replay.mode", "rllm.algorithm.router_replay"),
     ("actor_rollout_ref.rollout.n", "rllm.rollout.n"),
     ("actor_rollout_ref.rollout.val_kwargs.n", "rllm.rollout.n_val"),
+    ("actor_rollout_ref.rollout.temperature", "rllm.rollout.train.temperature"),
+    ("actor_rollout_ref.rollout.top_p", "rllm.rollout.train.top_p"),
+    ("actor_rollout_ref.rollout.top_k", "rllm.rollout.train.top_k"),
+    ("actor_rollout_ref.rollout.val_kwargs.temperature", "rllm.rollout.val.temperature"),
+    ("actor_rollout_ref.rollout.val_kwargs.top_p", "rllm.rollout.val.top_p"),
+    ("actor_rollout_ref.rollout.val_kwargs.top_k", "rllm.rollout.val.top_k"),
```
kylemontgomery1, line 62: We should  sync for max_tokens and data.max_response_length?

rllm/experimental/config/rllm/base.yaml
kylemontgomery1: what happens if i set both max_tokens and data.max_response_length? Which wins?

kylemontgomery1: The good thing about putting max_tokens here is that we can use different values for train/val.

But IMO we should not include max_tokens in the sampling params yet (just add it to the sampling params dict  using data.max_response_length) and defer this to when we refactor data to bake it backend agnositc.

rllm/engine/agentflow_engine.py
```diff
@@ -474,6 +479,7 @@ async def _run_single(self, task: dict | Task, uid: str, is_validation: bool = F
                     model=self.model,
                     session_uid=uid,
                     is_validation=is_validation,
```
kylemontgomery1, line 481: Is is_validation still needed, I think it's only used in the rollout engine to decide between the train and val sampling params?

cookbooks/agent_frameworks/agentflow/langgraph.py
```diff
@@ -32,7 +32,7 @@ async def langgraph_math(task: Task, config: AgentConfig) -> None:
         model=config.model,
         base_url=config.base_url,
         api_key="EMPTY",
-        temperature=1.0,
+        temperature=config.sampling_params.get("temperature", 1.0),
```
kylemontgomery1: Maybe it's better to encourage a pattern like **config.sampling_params?

rllm/engine/agentflow_engine.py
```diff
@@ -474,6 +479,7 @@ async def _run_single(self, task: dict | Task, uid: str, is_validation: bool = F
...
+                    sampling_params=(self.train_sampling_params if not is_validation else self.val_sampling_params) or {},
```
kylemontgomery1, line 482: Does remote agent flow need something similar?

kylemontgomery1:
I do like moving sampling params to the base config instead of keeping it backend specific. Unless there is an cli/evals reason for passing sampling params from the agent flow, I think we should discourage it; I think it's better to attach them to the rollout engine.

rllm/experimental/verl/utils.py
listar2000: As you mentioned above -- let's keep a single truth with `data.max_response_length` for now (and later remove this). One annoying thing for this `data` config group is that currently both `tinker.yaml` and `verl.yaml` sets this individually -- so it ought to be something backend-agnostic and elevated to `base.yaml` anyway.

rllm/engine/agentflow_engine.py
listar2000: The scope of this PR is just for `agentflow` now. For `remote_agentflow` I will let Sijun knows about this pattern and maybe integrate separately. 

One justification for this "delay" is that this PR can be seen as more as a feature improvement than bug fix (the user can still just modify the agent flow itself to change these params anyway).

listar2000:
@kylemontgomery1 Thanks for the review! For this PR I would like to just keep the original scope while taking your advice (so still having `data.max_response_length`). Will send a follow up later on moving this out as well.

rllm/experimental/config/rllm/base.yaml
listar2000: I think rollout and sampling are already similar?

listar2000: But maybe `rollout.sampling_params` makes sense.

listar2000: Actually I think it becomes very long after this additional layer, let's just keep this naming for now.

kylemontgomery1: Sure, at some point i'd also like to add rollout.chat_template_kwargs which could include things like disable_thinking, accumulate_reasoning, reasoning_effort, chat_template_path, etc. But I think this is a future pr.

cookbooks/agent_frameworks/agentflow/langgraph.py
listar2000: Good advice. I will then instead set these agentflow-specific params in their training scripts.

rllm/engine/agentflow_engine.py
listar2000: Keep this for now since it's already here -- might allow the agent flow to implement some different behavior depending on train/validation state.

PR #578, opened 2026-05-12
feat(eval): forward sampling params through `rllm eval` into AgentFlowEngine

listar2000:
> Stacked on top of #573 — review/merge that first; this PR's diff against `main` will shrink to just the eval/CLI changes once #573 lands.

## Summary

Wires sampling parameters through the `rllm eval` path so AgentFlows behave the same as in training, where #573 made flows spread `**config.sampling_params` into their LLM calls.

## Why

`rllm/eval/runner.py:run_dataset` constructs `AgentFlowEngine` **without** sampling params, so during eval `AgentConfig.sampling_params` is `{}`. After #573 (flows do `**config.sampling_params` instead of `.get(KEY, <local fallback>)`), eval LLM calls would silently fall back to the server's default sampling instead of anything the flow / user intended.

## What changed

- `run_dataset` gains a `sampling_params: dict | None = None` arg and threads it into `AgentFlowEngine` as both `train_sampling_params` and `val_sampling_params` (eval always runs as validation). Default is `{"temperature": 1.0, "top_p": 1.0}` — `max_tokens` is intentionally left unset so each flow keeps the server's own response budget rather than being clamped to a guessed value.
- `rllm eval` CLI gains `--temperature` / `--top-p` / `--max-tokens`, forwarded through `_run_eval` → `run_dataset`.

## Usage

```bash
rllm eval math --agent math --temperature 0.6 --max-tokens 8192
```

## Test plan

- [x] `pytest tests/cli/test_eval_command.py tests/eval/test_unified_runner.py tests/eval/test_eval_engine.py` — all pass
- [x] `ruff check` clean


listar2000:
Closed due to #623

PR #586, opened 2026-05-17
feat(agentcore): distinguish timeouts and propagate session_id to episode logs

luyuzhe111:
# Issue

When training with agentcore (e.g. `examples/agentcore_math/train_agentcore_math_verl.sh`), the wandb
`batch/*` panels and per-episode logs were missing key debugging information:

- `asyncio.TimeoutError` from `RolloutFuture.result_async` was conflated with every other transport failure, so `batch/termination_reason/timeout` stayed at 0 and all failures landed under `error`.
- `future.elapsed()` was only captured on the success path; transport errors recorded `elapsed=0.0`.
- Application-error detail (status_code, stop_reason, traceback from `@rollout_entrypoint`) was truncated.
- The AgentCore `session_id` (the only handle that resolves to a CloudWatch trace) was never written to the per-episode JSON, so `/check-agentcore-session-logs` could not link a failed episode back to its session.

# Summary

- Split `_run_one` so `asyncio.TimeoutError` produces `TerminationReason.TIMEOUT`, and capture `future.elapsed()` on every failure path.
- Preserve `status_code` / `stop_reason` / `traceback` from AgentCore application errors via `RemoteTaskResult.raw_result`, surface them in `episode.metadata.error`.
- Add typed `Episode.session_id` and write it to the per-episode JSON so `/check-agentcore-session-logs` can resolve a failed episode to its CloudWatch trace. Harbor's `metadata` use is unchanged.
- Restore `tests/engine/test_remote_runtime.py` (had bit-rotted: stale field names,missing imports) and add `.github/workflows/test-remote-runtime.yml` so it stays green.

# Changes

- `rllm/experimental/engine/remote_runtime/agentcore_runtime.py` — split submission vs polling exception handling; emit `TerminationReason.TIMEOUT` for`asyncio.TimeoutError`; ERROR-level logs for transport failures so they survive
  megatron's WARNING filter.
- `rllm/experimental/engine/remote_agent_flow_engine.py` — populate `episode.metadata["error"]` with `error_message`, `elapsed`, plus `stop_reason`/`traceback`/`status_code` from `result.raw_result` when available; reference the AgentCore `session_id` in the per-episode failure log line.
- `rllm/types.py` — add `Episode.session_id: str | None = None`; include in `Episode.to_dict()`.
- `rllm/experimental/engine/remote_agent_flow_engine.py` (`_build_episode`) — set `session_id` from `RemoteTaskResult.session_id`.
- `rllm/utils/episode_logger.py` — write top-level `session_id` and the full `metadata` dict to per-episode JSON.
- `tests/engine/test_remote_runtime.py` — fix imports/field renames; replace the single `result_async_raises` test with separate `TIMEOUT` and `ERROR` cases; assert `episode.session_id`.
- `.github/workflows/test-remote-runtime.yml` — new CI workflow guarding the test file and its production deps.

# Test plan

- [x] CI: `pytest tests/engine/test_remote_runtime.py` (16 tests) passes via `test-remote-runtime.yml`.
- [x] E2E `bash examples/agentcore_math/train_agentcore_math_verl.sh` with default `session_timeout=300`: wandb shows `batch/termination_reason/env_done ≈ 1`, `timeout ≈ 0`; per-episode JSON has top-level `session_id`, empty `metadata.error`.
- [x] E2E with override `rllm.remote_runtime.session_timeout=10`: `batch/termination_reason/timeout` populates; driver log shows `ERROR: Task <session> timed out after Ys` lines; per-episode JSON for timeout episodes carries `metadata.error.error_message="Timeout after Ys"`, `elapsed`, `termination_reason="timeout"`.

### DIFF
--- a/.github/workflows/test-remote-runtime.yml
+++ b/.github/workflows/test-remote-runtime.yml
@@ -1,0 +1,46 @@
+name: Test remote runtime
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - "rllm/experimental/engine/remote_runtime/**"
+      - "rllm/experimental/engine/remote_agent_flow_engine.py"
+      - "rllm/experimental/engine/gateway_manager.py"
+      - "rllm/experimental/engine/trace_converter.py"
+      - "rllm/utils/episode_logger.py"
+      - "rllm/types.py"
+      - "tests/engine/test_remote_runtime.py"
+      - ".github/workflows/test-remote-runtime.yml"
+  pull_request:
+    branches: [main]
+    paths:
+      - "rllm/experimental/engine/remote_runtime/**"
+      - "rllm/experimental/engine/remote_agent_flow_engine.py"
+      - "rllm/experimental/engine/gateway_manager.py"
+      - "rllm/experimental/engine/trace_converter.py"
+      - "rllm/utils/episode_logger.py"
+      - "rllm/types.py"
+      - "tests/engine/test_remote_runtime.py"
+      - ".github/workflows/test-remote-runtime.yml"
+  workflow_dispatch:
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Python 3.12
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v4
+
+      - name: Install dependencies
+        run: uv sync --extra agentcore --extra dev
+
+      - name: Run remote runtime tests
+        run: uv run pytest tests/engine/test_remote_runtime.py -v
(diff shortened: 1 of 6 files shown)


jeffreysijuntan:
Question: is the session_id only used by AgentCore, or do you intend it to be a more general field used by rllm and the gateway?

luyuzhe111:
@jeffreysijuntan Good question! For this PR, I scoped to agentcore only, but in the long term, I do think session id can be helpful for other engines as well, specifically for observability (otel), where it's become a standard to use session id as the root of trace spans. 

The semantics would be: unique prompt / task (task id) -> multiple rollouts (multiple session ids). And we can use session id for episode id, which currently has the format of task_id:rollout_idx. If I understand it correctly, we used rollout_idx more as a way to distinguish different rollouts, and the index value doesn't really carry much meanings?

And once we have a systematic structure (task id, session id, trace id, etc), we can probably think about exposing APIs for the user to transform complex traces arising from multi-agents / dynamic subagents for example.

jeffreysijuntan:
Got it. Lgtm.

luyuzhe111:
@jeffreysijuntan thanks for the review! I will go ahead and merge.

PR #596, opened 2026-05-26
feat(gateway): add a cumulative mode, if input OpenAI message list of one agent session is kept cumulative, user could choose to make model gateway construct cumulative input prompt token id sequence and directly input it to vllm server completion API

lyzustc:
## Summary

Adds optional `cumulative_token_mode` to the gateway. When enabled, the gateway will monitor per-session input OpenAI message text list, if it grows cumulatively (no context break like context compression happens), it will construct cumulative prompt token id sequence (by the [`renderers`](https://github.com/PrimeIntellect-ai/renderers) package) from turn 2, and directly send it to vllm completion API. This avoids retokenization drift --- in default mode, OpenAI message text list is directly input to vllm chat API, they are retokenized to prompt token id sequence every time.

If the message-list prefix diverges from the accumulated state at a turn, the accumulator resets and the request falls back to the normal chat path at that turn, and cumulation monitoring restarts at next turn.

Note that this mode is transparent to upstream agent framework. No need to change any code in agent framework. Agent framework could still send OpenAI chat request and receive OpenAI chat response normally. Input and output format transformation all happen within gateway.

## Why

Multi-turn RL training is sensitive to tokenization drift: re-tokenizing the message list each turn can produce a token sequence different from what the model originally generated. If users do not like it, they can choose to avoid it by using this mode.

## Changes

- **`token_accumulator.py`** — `TokenAccumulator` tracks per-session cumulative
    token state and delegates cross-turn bridging to the renderer's
    `bridge_to_next_turn(prev_prompt_ids, prev_completion_ids, new_messages, *, tools=None)`.
    `extract_new_messages` slices the unseen tail of the message list and drops
    assistant turns (the model re-samples those). Truncated completions are closed
    with a synthesized end token by the renderer; assistant messages in the new slice
    cause the bridge to bail out (falling back to the chat path).
- **`proxy.py`** — `ReverseProxy` gains `cumulative_token_mode` and `renderer`. In
    `handle()`, chat-completions requests on an established session are rewritten to
    `/v1/completions` with the cumulative prompt; if the bridge returns `None`, it
    falls back to the normal chat path.
- **`data_process.py`** — extractors handle the `/v1/completions` response shape
    (root-level `prompt_token_ids`, flat `token_logprobs`).
- **`server.py`** — when `cumulative_token_mode=True`, loads the tokenizer from
    `config.model` and builds a renderer via `create_renderer(tokenizer, renderer=config.renderer_family)`.
    Raises a hard startup error if this resolves to `DefaultRenderer` (which provides
    no cross-turn bridge). New CLI flags: `--cumulative-token-mode`, `--renderer-family`.
- **`models.py`** — `GatewayConfig` gains `cumulative_token_mode: bool` and
    `renderer_family: str = "auto"`.
- **`pyproject.toml`** — adds `renderers>=0.1.7` and `transformers>=4.50.0`.

## Example usage

```bash
python -m rllm_model_gateway.server \
    --worker http://localhost:8000/v1 \
    --model Qwen/Qwen3-8B \
    --cumulative-token-mode \
    --renderer-family qwen3
```
In RL training, set in `gateway:` config:

```yaml
gateway:
  cumulative_token_mode: true
  renderer_family: qwen3   # or "auto" if model is a canonical HF id
```

## Tests

- [x] pytest rllm-model-gateway/tests/unit/
- [x] Multi-turn rollout vs real vLLM; verify turn-N prompt id sequence prefix equal turn-(N-1) prompt + completion + environment observation.
- [x] Verify reset path triggers on context compression.
- [x] cumulative_token_mode=False (default) behavior unchanged.
- [x] Test it in a real agentic RL experiment (on AppWorld) agent and ensure cumulative input token sequence is maintained.

### DIFF
--- a/rllm-model-gateway/pyproject.toml
+++ b/rllm-model-gateway/pyproject.toml
@@ -16,6 +16,8 @@ dependencies = [
     "pydantic>=2.0.0",
     "aiosqlite>=0.19.0",
     "PyYAML>=6.0",
+    "renderers>=0.1.7",
+    "transformers>=4.50.0",
 ]
 
 [project.optional-dependencies]
--- a/rllm-model-gateway/src/rllm_model_gateway/data_process.py
+++ b/rllm-model-gateway/src/rllm_model_gateway/data_process.py
@@ -21,11 +21,17 @@ logger = logging.getLogger(__name__)
 
 
 def extract_prompt_token_ids(response: dict[str, Any]) -> list[int]:
-    """Extract ``prompt_token_ids`` from the root of a vLLM response."""
+    """Extract ``prompt_token_ids`` from a vLLM response.
+
+    Checks root level first (chat/completions format), then falls back to
+    choices[0].prompt_token_ids (completions format).
+    """
     ids = response.get("prompt_token_ids")
     if ids is None:
-        return []
-    return list(ids)
+        choices = response.get("choices")
+        if choices:
+            ids = choices[0].get("prompt_token_ids")
+    return list(ids) if ids is not None else []
 
 
 def extract_completion_token_ids(response: dict[str, Any]) -> list[int]:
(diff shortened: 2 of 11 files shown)


jeffreysijuntan:
Nice feature! Can we turn this behavior on by default?

lyzustc:
> Nice feature! Can we turn this behavior on by default?

I am ok with that. But want to hear what do you think @luyuzhe111 ?

luyuzhe111:
@lyzustc @jeffreysijuntan maybe keep it off by default for now? I can run some e2e experiments to stress test it, after which we can turn it on by default?

lyzustc:
> @lyzustc @jeffreysijuntan maybe keep it off by default for now? I can run some e2e experiments to stress test it, after which we can turn it on by default?

That sounds good to me.

kylemontgomery1:
Yeah I agree, probably leave disabled by default for now. I'm running a quick test with this feature on agentcore math.

PR #615, opened 2026-06-05
feat(eval): add Tinker as an OpenAI-compatible eval provider

listar2000:
## Why

[Tinker](https://thinkingmachines.ai/) (Thinking Machines) is one of rLLM's two training backends — but until now you could only *train* with it, not *evaluate* against it through the standard `rllm eval` CLI. Tinker recently shipped a (beta) **OpenAI-compatible inference endpoint**, which lets us treat it as just another provider — no native `SamplingClient`, no bespoke plumbing. This closes the loop: configure Tinker once and use the same `rllm model setup` → `rllm eval` flow you use for OpenAI/Anthropic/Together/etc.

## What's supported

- **First-class `tinker` provider** in `rllm model setup` / `swap` / `show`, authenticated with `TINKER_API_KEY`.
- **Live model catalog** — `setup` pulls the server's supported models from `ServiceClient.get_server_capabilities()` (41 models today: Qwen3, Llama, DeepSeek, Kimi, Nemotron, gpt-oss, …), with a curated static fallback when offline / unkeyed.
- **Base models *and* fine-tunes** — the OAI endpoint accepts:
  - a **base model ID** like `Qwen/Qwen3-8B` (no checkpoint / training required), or
  - a **`tinker://…/sampler_weights/…` checkpoint path** (pasted via "Other (enter manually)") to eval a fine-tune.
- **Consistent routing** — goes through the existing LiteLLM proxy → gateway path like every other provider; Tinker just pins LiteLLM's `openai/` adapter to a fixed `api_base`.

## How it works

`ProviderInfo` gains an optional `base_url`. `EvalProxyManager` injects that as the LiteLLM `api_base` only for providers that declare one (so OpenAI/Together/etc. are byte-for-byte unchanged). Generated LiteLLM params for Tinker:

```yaml
model: openai/Qwen/Qwen3-8B
api_base: https://tinker.thinkingmachines.dev/services/tinker-prod/oai/api/v1
api_key: $TINKER_API_KEY
```

## Testing

End-to-end through the real endpoint:

```
rllm eval gsm8k --split test --max-examples 3   # Qwen/Qwen3-4B-Instruct-2507
→ Accuracy 100.0% (3/3), 0 errors
```

Also verified: live catalog fetch returns 41 models; `api_base` is injected for Tinker only (no regression for other providers); `ruff` clean.

## Files

- `rllm/eval/config.py` — `tinker` provider, `ProviderInfo.base_url`, `fetch_tinker_models()` live-catalog helper.
- `rllm/eval/proxy.py` — inject `api_base` for fixed-URL providers.
- `rllm/cli/_ui.py`, `rllm/cli/model_cmd.py` — live catalog in model picker; thread API key through.
- `docs/quickstart-cli.mdx` — Tinker setup note.

### DIFF
--- a/docs/quickstart-cli.mdx
+++ b/docs/quickstart-cli.mdx
@@ -31,6 +31,10 @@ You'll be prompted to:
 Your configuration is saved to `~/.rllm/config.json`. You can switch providers later with `rllm model swap`.
 </Info>
 
+<Tip>
+Selecting **Tinker (Thinking Machines)** evaluates against Tinker's (beta) OpenAI-compatible endpoint using your `TINKER_API_KEY`. The model list is fetched live from the Tinker server — pick any base model (e.g. `Qwen/Qwen3-8B`, no checkpoint needed) or paste a `tinker://…` sampler checkpoint path to evaluate a fine-tune.
+</Tip>
+
 ## Step 2: Explore available datasets
 
 Browse the full catalog of 50+ benchmarks:
--- a/rllm/cli/_ui.py
+++ b/rllm/cli/_ui.py
@@ -103,9 +103,24 @@ def _select_provider(existing: RllmConfig) -> str:
     return provider_ids[idx]
 
 
-def _select_model(provider: str, existing: RllmConfig) -> str:
-    """Interactive model selection with option to enter a custom model."""
+def _select_model(provider: str, existing: RllmConfig, api_key: str | None = None) -> str:
+    """Interactive model selection with option to enter a custom model.
+
+    For Tinker, the model list is pulled live from the server's capabilities
+    (``api_key`` lets the fetch authenticate); the curated list is the fallback.
+    """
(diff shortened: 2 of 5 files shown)


jeffreysijuntan:
There should be more models available from the Tinker side? Can we add them all?

listar2000:
> There should be more models available from the Tinker side? Can we add them all?

Look into this. Some models have been deprecated though.

listar2000:
@jeffreysijuntan I think what you've seen is the fall-back model catalog. Since Tinker-supported models change dynamically, we will live-fetch the model list (which requires the API key in the first place). 

I've updated the logic to use `.env` TINKER_API_KEY by default if available.

b48d5959-15c4-493e-86cd-b22cdeed0cc4

jeffreysijuntan:
oh I see, make sense

PR #616, opened 2026-06-05
fix(eval): close sandbox in hook teardown so cloud sandboxes get deleted

jeffreysijuntan:
## Summary

- `SandboxTaskHooks` creates the per-task sandbox and injects it into a `SandboxedAgentFlow` via `set_sandbox()`, which marks the sandbox as externally-managed. The hook's teardown then called `task_flow.teardown_sandbox()`, which is a **no-op** for externally-managed sandboxes — so `sandbox.close()` was never called and the Daytona delete API never fired.
- Net effect: every `rllm eval` task on the Daytona backend left an orphaned cloud sandbox. The 30-min `auto_stop_interval` would eventually stop it, but no `auto_delete_interval` is set, so the sandbox stayed in the user's account.
- Fix: the hook owns the sandbox lifecycle (it created it), so close it directly from the teardown closure. Same fix benefits Modal / e2b / runloop, which also bill by duration.

## Test plan

- [x] New regression test `test_hooks_close_sandbox_for_sandboxed_agent_flow` in `tests/eval/test_unified_runner.py` asserts that the hook teardown calls `sandbox.close()` exactly once when the agent flow is a `SandboxedAgentFlow` (previously failed: 0 calls).
- [x] `uv run pytest tests/eval/` — 400 passed (1 pre-existing failure in `test_eval_engine_populates_steps_from_gateway_traces`, unrelated to this change: `_FakeGateway` is missing `adelete_sessions`).
- [x] Manual smoke: run `rllm eval <bench> --agent bash --sandbox-backend daytona --max-examples 2` and confirm no orphaned sandboxes remain in the Daytona dashboard after completion.

### DIFF
--- a/rllm/cli/eval.py
+++ b/rllm/cli/eval.py
@@ -219,8 +219,14 @@ def _run_eval(
             else:
                 split = "test"
 
-        # Docker check for Harbor tasks
-        if (agent_name and agent_name.startswith("harbor:")) or (catalog_entry and catalog_entry.get("source", "").startswith("harbor:")):
+        _is_harbor_agent = bool(agent_name) and agent_name.startswith("harbor:")
+        _is_harbor_source = bool(catalog_entry) and catalog_entry.get("source", "").startswith("harbor:")
+
+        # Require Docker only when execution lands on the local daemon; a remote
+        # backend (modal/daytona) pulls the task's image in the cloud.
+        _eff_backend = (agent_metadata or {}).get("sandbox_backend")
+        _runs_on_local_docker = _eff_backend in (None, "docker")
+        if (_is_harbor_agent or _is_harbor_source) and _runs_on_local_docker:
             from rllm.integrations.harbor.utils import diagnose_docker
 
             ok, reason, hint = diagnose_docker()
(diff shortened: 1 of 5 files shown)


listar2000:
### Follow-up commit `65c044ed`: make the daytona path runnable + cleanup observable

Two fixes needed to actually exercise this PR end-to-end on a remote backend:

**1. Unblock remote harbor eval.** The harbor Docker precondition required a local Docker daemon for *any* harbor dataset, so `rllm eval harbor:<bench> --sandbox-backend daytona` failed before a sandbox was ever created — even though daytona/modal pull the image in the cloud. Scoped the gate to local-docker backends and routed harbor-source rows past on-the-fly materialise + the dataset-wide-evaluator requirement (each task carries its own `tests/test.sh`). Cherry-picked from #617, so expect a trivial `eval.py` overlap at merge.

**2. Surface delete failures.** `DaytonaSandbox.close()` swallowed `delete()` errors at debug level — a failed delete (e.g. an API key without delete scope → `Access denied`) leaks silently, defeating the point of this PR. Now logs at WARNING and best-effort `stop()` so compute halts immediately instead of at the 30-min `auto_stop`.

The core teardown fix is correct as-is; these just make it reachable and its failures visible.

PR #617, opened 2026-06-06
fix(eval): run harbor SWE-bench via the native mini-swe-agent harness on remote sandboxes

listar2000:
## What

Makes this work end-to-end:

```bash
rllm eval harbor:swebench-verified --agent mini-swe-agent --sandbox-backend modal
rllm eval harbor:swebenchpro      --agent mini-swe-agent --sandbox-backend daytona
# provider: rllm model show → OpenRouter / anthropic/claude-sonnet-4-6
```

i.e. evaluate a Harbor SWE-bench dataset with a **built-in** rLLM harness on a **remote** cloud sandbox (Modal/Daytona), scored by rLLM's own per-task verifier — no Harbor `swerex`/trial runtime, no bespoke SWE module.

## Why it didn't work before

A Harbor task-per-directory dataset already ships everything rLLM needs (`environment/Dockerfile` + `tests/test.sh`), and `ShellScriptEvaluator` is already Harbor-convention-aware (`/tests`, `/logs/verifier/reward.txt`). Harbor's own `harbor_reward_fn` only reads a reward that `HarborRuntime` pre-computes, so a **native** agent needs the per-task `tests/test.sh` verifier instead — which rLLM already falls back to. The missing piece was wiring each task's **image / workdir / verifier / resources** through to a remote sandbox.

**Resolution wiring (image / workdir / verifier):**

- **Per-task image never reached the sandbox.** The SWE-bench image is declared as `FROM <registry>/<instance>` in the task's `environment/Dockerfile`; it wasn't surfaced anywhere, so non-docker backends fell back to `python:3.11-slim` (no repo). → parse the Dockerfile `FROM`/`WORKDIR` into task metadata (`docker_image` + workdir).
- **Eval flattened Harbor task-dirs.** For a non-Harbor agent, eval tried to materialise the rows into a flat "simple" benchmark (dropping `tests/` + `environment/`) and required an explicit evaluator. → Harbor-source datasets now skip materialise, wrap rows with `task.toml` merge, and allow the per-task `tests/test.sh` verifier (no dataset-wide evaluator).
- **Docker precondition blocked Modal.** A working local Docker daemon was required for *any* Harbor dataset, even with `--sandbox-backend modal`. → gate only when the effective backend is local docker.
- **`create_sandbox` dropped the image for Modal (latent bug).** It passed `image=` to docker/daytona but not Modal, so **every** Modal sandbox silently used `python:3.11-slim`. → pass it through.
- **Base image lacks the verifier's toolchain.** Pulling the Dockerfile's `FROM` base (instead of building it) skips its `RUN` steps — including `install uv`, which the swebench grader needs. Without it the tests ran but the grader didn't, so even a golden patch scored 0. → replay the Dockerfile's `RUN` steps on non-docker sandboxes.

**Sandbox plumbing (backend-generic; surfaced live on Daytona, latent on Modal):**

- **Uploaded dirs carried host uid/gid.** `upload_dir` tarred with the host's uid/gid; root extraction in the sandbox tried to `chown` to nonexistent ids and exited 2 — a hard crash on Daytona (raising `exec`), silently swallowed on Modal. → extract with `tar --no-same-owner` (permissions kept, so `test.sh` stays +x).
- **Declared task resources were ignored.** `_create_sandbox_for_task` never passed the `[environment]` resources task.toml declares (`memory_mb`, `cpus`, …), so remote sandboxes ran at the backend default (Daytona: **1 GiB**). Compile-heavy graders (`go test ./...`, Go not cgroup-CPU-aware → 64 parallel compiles) OOM-killed the build → 0 tests pass → even the golden patch scored 0. → honor declared cpus/memory/storage per backend (Modal MB, Daytona GB).

All changes live in the shared eval/sandbox path; no SWE-specific module.

## Validation (real cloud runs, OpenRouter / Sonnet 4.6)

- **Modal · swebench-verified · oracle** (golden patch + verifier, no LLM): **2/3 reward 1.0**. The miss returns FAILED from Harbor's *verbatim* `parser.py` with all 242 tests passing — a benchmark grading quirk for that instance, reproducible independent of rLLM.
- **Modal · swebench-verified · mini-swe-agent**: **2/2 = 100%** on the gradeable tasks — the agent genuinely fixed both real GitHub issues (31 / 55 LLM steps).
- **Daytona · swebenchpro** (different backend *and* dataset — generality check): before the two sandbox fixes, both tasks errored (tar/chown crash). After: **Errors=0** with real verifier verdicts (`setup=8s evaluator=32s`), and the Go-grader oracle that OOM'd at the 1 GiB default runs cleanly at the honored 4 GiB (`oom_kill 0`).

So failures are attributable to genuine agent mistakes or the benchmark's own grading — not harness/environment plumbing — across two backends and two datasets.

## Files

- `rllm/tasks/loader.py` — `_parse_dockerfile_image_workdir` + Dockerfile parsing in `_merge_task_toml_metadata`.
- `rllm/cli/eval.py` — `_dict_rows_to_tasks` merges `task.toml`; Harbor-source skips materialise + allows per-task verifier; docker gate scoped to local-docker.
- `rllm/sandbox/sandboxed_flow.py` — pass `image` to the Modal backend.
- `rllm/eval/_resolution.py` — `_dockerfile_run_commands` replay on non-docker sandboxes; `_sandbox_resource_kwargs` honors task.toml resources.
- `rllm/sandbox/backends/{daytona,modal_backend}.py` — `tar --no-same-owner` on `upload_dir`.

### DIFF
--- a/rllm/cli/eval.py
+++ b/rllm/cli/eval.py
@@ -37,13 +37,15 @@ def _suggest_benchmarks(name: str, catalog_names: list[str], max_suggestions: in
 def _dict_rows_to_tasks(rows: list[dict]) -> list:
     """Wrap dict-rows from a catalog dataset as Task objects.
 
-    Harbor rows carry ``task_path`` pointing at their own Harbor task
-    directory; we root the Task there so HarborRuntime and per-task
-    verifier resolution find ``task.toml`` / ``tests/`` natively. Other
-    rows get ``dataset_dir=Path(".")`` and rely on ``evaluator_override``.
+    Harbor rows carry ``task_path``; we root the Task there and merge the task's
+    ``task.toml`` + ``environment/Dockerfile`` metadata (workdir, verifier
+    timeout, per-task image) so per-task verifier resolution and an rllm-native
+    harness get the right sandbox. Other rows get ``dataset_dir=Path(".")`` and
+    rely on ``evaluator_override``.
     """
     from pathlib import Path
 
+    from rllm.tasks.loader import _merge_task_toml_metadata
     from rllm.types import Task
 
     tasks: list[Task] = []
@@ -52,11 +54,14 @@ def _dict_rows_to_tasks(rows: list[dict]) -> list:
         task_id = str(row.get("id") or row.get("task_id") or idx)
         task_path = row.get("task_path")
         dataset_dir = Path(task_path) if task_path else Path(".")
+        metadata = dict(row)
+        if task_path:
+            metadata = _merge_task_toml_metadata(dataset_dir, metadata)
         tasks.append(
             Task(
                 id=task_id,
                 instruction=str(instruction),
-                metadata=dict(row),
+                metadata=metadata,
                 dataset_dir=dataset_dir,
                 sub_dir=None,
             )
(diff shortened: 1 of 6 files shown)


listar2000:
@jeffreysijuntan Would appreciate your review here -- a bit worried whether these changes (not a lot though) will cause any side-effect.

PR #621, opened 2026-06-06
fix(gateway): add plumbing for weight_version and routing matrices

kylemontgomery1:
## Summary

The model gateway stamped traces with its own `proxy.weight_version` and ignored the version the rollout engine actually produced, and it had no path at all for R3 routing matrices. This adds plumbing so per-rollout `weight_version` and routing matrices flow from the rollout-engine response through the gateway into the trace (preferring the response value, falling back to the gateway stamp for vLLM, which emits neither). It also guards `verl + gateway + router_replay=R3`, a combination that would silently drop routing because verl captures routed experts off the worker, not over the gateway's HTTP path.

## Type of change

- [ ] Feature
- [x] Fix
- [ ] Docs
- [ ] Refactor
- [ ] Example / Project
- [ ] Infra / CI

## What changed

- Gateway now takes per-rollout `weight_version` from the rollout-engine response when present, falling back to the gateway-tracked version for vLLM.
- R3 routing matrices flow end-to-end through the gateway (adapter → `extract_routing_matrices` → `TraceRecord` → trace_converter → `ModelOutput`/`Step`); both fields are stripped from the client-facing response.
- `ModelOutput.to_dict`/`from_dict` now serialize `routing_matrices` so it survives episode offload/reload in the async buffer.
- Guard at trainer init: `verl + gateway (agent_flow/remote_runtime) + router_replay=R3` raises; R2 and non-gateway paths are unaffected.

## Validation

- [x] Targeted tests: `pytest rllm-model-gateway/tests/unit/test_data_process.py tests/engine/test_trace_converter.py tests/engine/test_tinker_engine.py`
- [ ] Not run (reason below)

Validation details:
- ruff + ruff-format ran via the commit hook and passed.
- 42 targeted tests passed (data_process, trace_converter, tinker_engine).
- No end-to-end training run performed; `weight_version` is metrics-only (`async/staleness_*`) and never gates or drops a group.

- None

## Docs / examples

- [x] Not needed
- [ ] Updated docs
- [ ] Updated examples
- [ ] Follow-up docs needed

## Related issues / PRs

- Fixes #
- Related to #
- Stacked on / depends on #

## Screenshots / logs

- N/A


listar2000:
LGTM

PR #623, opened 2026-06-07
feat(sampling): gateway-enforced sampling-parameters config for AgentFlow

listar2000:
## Summary

Wires sampling parameters into AgentFlow with **one rule**: sampling config is resolved at the launcher (CLI flags / script overrides) into a single `SamplingConfig`, attached to the gateway session, and the **gateway overwrites those keys on every outgoing LLM request** — whatever the agent program or harness put there. A key is owned by exactly one place: if it's in the resolved launcher config the gateway owns it and always wins; if it's absent, the flow/harness keeps it. Same behavior for whitebox (`@rllm.rollout`) and blackbox (Claude Code, mini-swe-agent) programs, because enforcement happens at the proxy every call already routes through.

Supersedes #578 (its three `rllm eval` flags become a subset of `--sampling-params`, now routed through the gateway session so they also bind blackbox harnesses).

## What changed

- **Schema + resolver** — `rllm/cli/_sampling.py`: a small `SamplingConfig` (typed core `temperature`/`top_p`/`top_k`/`max_tokens` + an open `extra` passthrough for `presence_penalty`, `min_p`, …) plus the CLI resolvers. `--sampling-params` accepts a `"k=v,…"` string or `@file.yaml`/`@file.json`; precedence is `base.yaml rollout.{train,val}` < `@file`/string < `--temperature`/`--top-p`/`--max-tokens` flags. Resolved params are forwarded to the backend as-is — no provider-specific filtering; an unsupported key fails at the backend (the user's responsibility).
- **Defaults** — `base.yaml rollout.{train,val}` defaults only the universally accepted params (`temperature`/`top_p`/`max_tokens`); `top_k` (and penalties, etc.) are opt-in, so nothing non-portable is auto-injected. Native verl keeps its own `top_k` default via `actor_rollout_ref.rollout.*`.
- **Gateway enforcement** — `_mutate` simplifies to `payload.update(sp)` (configured keys win). Removes the `sampling_params_priority` knob entirely (middleware, `GatewayConfig`, server CLI, `base.yaml`, `GatewayManager`).
- **CLI** — `--sampling-params` + `--temperature`/`--top-p`/`--max-tokens` on both `rllm eval` and `rllm train`. The local `AgentFlowEngine` attaches the resolved config to the gateway session, so **whitebox flows are gateway-enforced too** (previously only blackbox harnesses were).
- **Flow demotion** — `AgentConfig.sampling_params` is now an informational mirror; the 8 cookbook flows drop the `**config.sampling_params` spread + manual `top_k` strip and just make plain LLM calls (the gateway injects server-side).

## CLI examples

```bash
# eval — gateway pins temperature + presence_penalty for every call the agent makes
rllm eval math --agent math --temperature 0.6 --sampling-params "presence_penalty=0.1"

# eval of a blackbox harness that hard-wires temperature — gateway overrides it
rllm eval swe-bench --agent harbor:mini-swe-agent --sampling-params "temperature=0.2,top_p=0.9"

# AgentFlow training (Hydra) — extra keys via the rllm.rollout namespace
python train.py rllm/backend=tinker \
  rllm.rollout.train.temperature=0.6 +rllm.rollout.train.presence_penalty=0.1
```

## Breaking changes

- Conflict resolution is **gateway-always-wins** for configured keys (the old middleware default let per-call client values win). Intended; release-note it.
- `gateway.sampling_params_priority` removed from `base.yaml` and the middleware.
- `AgentConfig.sampling_params` demoted to an informational mirror (cookbooks no longer spread it).
- `base.yaml` no longer ships `top_k: -1` under `rollout.{train,val}` — it's opt-in now. Native verl is unaffected (keeps its own default); AgentFlow won't inject `top_k` unless you set it.

## Scope

Scoped to the AgentFlow + gateway path. Native (non-AgentFlow) verl rollout is untouched — it already reads `temperature`/`top_p`/`top_k` from `actor_rollout_ref.rollout.*`, mirrored from `rllm.rollout.*` by `sync_config`.

## Testing

- **Unit**: `SamplingConfig` schema + CLI resolver precedence; gateway middleware always-wins.
- **Integration / E2E**: real gateway + mock upstream enforcement (override / passthrough / extra-key injection); eval `run_dataset` and the training-rollout path through a real gateway.
- **Real Tinker run**: 2-step LoRA training; forwarded requests carried injected `temperature=0.123` + `presence_penalty=0.5` on every rollout call.

### DIFF
--- a/rllm/cli/eval.py
+++ b/rllm/cli/eval.py
@@ -21,6 +21,7 @@ from rich.theme import Theme
 
 from rllm import paths
 from rllm.cli._pull import load_dataset_catalog, pull_dataset
+from rllm.cli._sampling import SAMPLING_PARAMS_HELP as _SAMPLING_PARAMS_HELP
 
 logger = logging.getLogger(__name__)
 
@@ -86,6 +87,7 @@ def _run_eval(
     save_episodes: bool = True,
     episodes_dir: str | None = None,
     use_snapshot: bool = True,
+    sampling_config=None,
 ):
     """Core eval logic, extracted for clean proxy lifecycle management."""
     from rllm.data import DatasetRegistry
@@ -382,6 +384,8 @@ def _run_eval(
     table.add_row("Evaluator", f"[dim]{evaluator_display}[/]")
     if not use_snapshot:
         table.add_row("Snapshots", "[dim]disabled (--no-snapshot, cold start)[/]")
+    if sampling_config is not None and not sampling_config.is_empty:
+        table.add_row("Sampling", f"[dim]{sampling_config.as_dict()} (gateway-enforced)[/]")
     console.print()
     console.print(Panel(table, border_style="cyan", expand=False))
     console.print()
@@ -487,6 +491,7 @@ def _run_eval(
             dataset_name=getattr(dataset, "name", benchmark) or benchmark,
             on_episode_complete=on_episode_complete,
             evaluator_override=evaluator,
+            sampling_params=(sampling_config.as_dict() if sampling_config is not None else None),
         )
     )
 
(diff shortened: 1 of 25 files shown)


listar2000:
@kylemontgomery1 This is a PR for what we said earlier: enforcing the sampling params at the gateway level instead of the agentflow level (esp. for many external harnesses we do not have explicit control over how they make API calls). I still leave explicit flags (or fields in `base.yaml`) for important params like `temperature`/`top_k`. But extra params like "presence_penalty=0.1" can also be specified via a string/yaml-based way.

kylemontgomery1:
Thanks! Will take a look and get this merged today.

rllm/cli/eval.py
```diff
@@ -85,8 +86,15 @@ def _run_eval(
     enable_ui: bool = False,
     save_episodes: bool = True,
     episodes_dir: str | None = None,
+    sampling_config=None,
+    sampling_capability: str | None = None,
```
kylemontgomery1: What is `sampling_capability` for? I think it's on the user if they pass an invalid sampling param; we should let it fail instead of silently dropping it?

listar2000: Yeah makes sense. `capability` is kind of a bad naming and we have to enforce a set of rules for different message formats. Agree with your suggestion.

PR #630, opened 2026-06-08
feat(trainer): deterministic optimizer steps with decoupled batch sizes to deal with variable seqs

luyuzhe111:
### Summary

- Decouple `mini_batch_size` (rows per optimizer step) from `global_batch_size` (seq-mean loss denominator) in the verl actor update path. Multi-turn rollouts produce a variable row count N that drifts across steps; the old code tied both sizes to `ppo_mini_batch_size * rollout_n`, which could cause excessive padding or variable number of optimizer updates, which makes it challenging to control the optimization process. 
- The number of optimizer steps per generation batch is now a deterministic `r = train_batch_size // ppo_mini_batch_size`, with padding minimized to the smallest multiple of `dp_size` that yields exactly `r` chunks.
- The loss denominator (`global_batch_size`) is fixed at `total_rollouts // r`, preserving conventional per-example loss scale regardless of row count drift.
- Padding is reduced from `lcm(dp_size, ppo_mbs * rollout_n)` to `ceil(N/r)` rounded to `dp_size`. For example, with `dp_size=2, ppo_mbs=32, rollout_n=4, N=298`: old divisor = `lcm(2, 128) = 128`, padding 298 to 384 (86 wasted rows); new divisor = 150, padding 298 to 300 (2 wasted rows).

### Test plan

Tested on agentcore-math (Qwen3-4B, GRPO, megatron, LoRA) with two configurations:

- **TP=8, dp=1, r=1** (`ppo_mini_batch_size=64`): almost full training, stable reward curve, `num_updates=1` and `global_batch_rollouts=256` every step.
- **TP=4, dp=2, r=2** (`ppo_mini_batch_size=32`): 22 steps (early terminated), `num_updates=2` and `global_batch_rollouts=128` every step, padding correctly rounds to multiples of 2 (e.g. 298->300, 289->292).

We can see that with this PR, the training performance still closely tracks the previous run (grey curve). But this should give a more consistent behavior that makes tuning easier to reason. The padding waste is also greatly reduced as mentioned in the example above.

### Limitation
- Multi-agent training use case was NOT covered. 

Screenshot 2026-06-08 at 12 32 53 PM

### DIFF
--- a/rllm/trainer/verl/verl_backend.py
+++ b/rllm/trainer/verl/verl_backend.py
@@ -497,24 +497,69 @@ class VerlBackend(BackendProtocol[Iterable, DataProto]):
         if dp_size is None:
             return batch
 
-        # From verl RayPPOTrainer._update_actor: ppo_mini_batch_size is multiplied
-        # by rollout.n before being passed as mini_batch_size to update_actor and
-        # make_iterator enforces batch_per_gpu % mini_batch_size == 0. Globally this
-        # requires batch_size % (ppo_mini_batch_size * rollout.n) == 0.
+        # Multi-turn rollouts are prefix-merged into training rows on a best-effort
+        # basis; the merge frequently breaks (chat-template re-render, context
+        # management), so the real row count N is much larger than the rollout
+        # count and varies step-to-step. We make the number of optimizer steps a
+        # deterministic quantity r derived from config, and decouple the two batch
+        # sizes the verl worker consumes (which it treats independently):
+        #   - mini_batch_size = m  (ROWS per optimizer step) controls how
+        #     train_mini_batch chunks the batch, hence the update count. Verl
+        #     requires m % dp_size == 0 and total_rows % m == 0.
+        #   - global_batch_size = gbs (ROLLOUTS per update) is the seq-mean loss
+        #     denominator. Keeping it at the constant rollout count R_total // r
+        #     (not the drifting row count) gives the conventional per-example loss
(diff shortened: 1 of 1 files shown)


listar2000:
Thanks @luyuzhe111 for the great work! I've went through the math a bit and these are correct (great improvements!). And I will probably do a follow-up PR on the **multi-agent** path soon. Just two minor comments (might be wrong):

**1. The clean `gbs = ppo_mbs * rollout_n` identity depends on `train_batch_size % ppo_mini_batch_size == 0`.** If it's not divisible, `r` gets floored and `gbs` drifts off the conventional value, subtly changing the seq-mean scale. Not sure if there's already a `verl`-level assert/warning guarding against this.

**2. Minor: `actor/num_updates` undercounts when `ppo_epochs > 1`.** `make_iterator` loops over epochs, so real optimizer steps = `r * ppo_epochs`, but the metric reports `r`. Harmless at the default `ppo_epochs=1` — but maybe worth multiplying through or renaming sth as "updates per epoch"? What do you think.

luyuzhe111:
@listar2000 thanks for the review! regarding your questions:
- right so this situation only arises only if someone has a "weird" config. say they specify train_batch_size = 64 but mini_ppo_batch_size to be 48. Not sure what's the intended behavior in this case regardless what's the number of seqs from the multi-turn rollout. Maybe it's worth a verl-level warning in a separate PR in this case?
- that's a good catch! just updated to `actor/num_updates_per_epoch`.

listar2000:
Thanks for the update!

PR #636, opened 2026-06-10
refactor(hooks,engine): one env-requirement join + always-hooks engine (AgentFlow phases 0–1)

listar2000:
## Intuition

> See https://github.com/rllm-org/rllm/discussions/639 for design doc.

Whether a rollout needs a sandbox used to be answered by **three different formulas in three places** (`needs_sandbox_isolation`, an `isinstance` check inside `SandboxTaskHooks.setup`, and `_needs_sandbox` in eval resolution) — and they disagreed. Each component now **declares** what it needs, and one pure function joins the declarations:

```
needs_env = flow.needs_env  ∨  env-style verifier (sandbox-shell / python-hybrid)  ∨  task declares an environment
```

- **Per task** (`resolve_rollout_plan`): decides sandbox + verifier at bind time. Includes the *no-consumer rule*: a task-declared env that neither the flow nor the verifier can reach is downgraded with a warning instead of provisioned for nobody.
- **Per run** (`scan_env_requirements`): the same predicate over **all** dataset rows (the old probe only looked at the first), deciding whether to install hooks and whether the gateway needs a public tunnel.
- **Declarations instead of isinstance**: `needs_env` (`SandboxedAgentFlow=True`, `@rllm.rollout=False`) and `llm_inside_env` (`BaseCliHarness=True` → in-sandbox process gets the public/tunnel URL; host-side loops keep the local URL so they never depend on a quick-tunnel hostname that can rotate mid-run).
- **Always-hooks engine**: evaluation policies `FixedEvaluation` / `FromTaskEvaluation` replace `evaluator_override` threading; a bare `evaluator=` wraps into `FixedEvaluatorHooks`, so `AgentFlowEngine` has exactly one execution path (the `ctx is None` forks and dict-vs-Task calling-convention split are deleted).

## Bug fixes shipped with the join

- **Training ran with strict trace enrichment silently off** for sandboxed flows (`strict = hooks is None`, but sandboxed training always has hooks) — empty token IDs could reach loss math. Now `strict = not is_validation`.
- Gateway loopback pin + tunnel decision now also apply when `hooks=` is passed explicitly (previously skipped entirely).
- `CloudflaredTunnel` blocks until the hostname actually answers over HTTP (`RLLM_TUNNEL_REACHABLE_TIMEOUT_S`, default 120 s) — cloudflared printing the URL only means it was *assigned*; rollouts burned all retries on the DNS race in ~4 s.
- `rllm train` no longer hard-fails on `sandbox-shell` / `python-hybrid` verifiers — the per-task policy resolves them in-sandbox, enabling **harbor-style graded training from the CLI for the first time**.

## Verification

- `tests/eval/test_env_join.py` (new): the join, the no-consumer rule, the all-rows scan, tunnel precedence.
- Full suite vs pristine HEAD: 0 regressions, 2 previously-broken test modules fixed (`test_eval_engine`, `test_agentflow_engine`), 14 shared pre-existing failures untouched (verl import drift etc.).
- **E2E on Tinker + Daytona** (snapshot + warm pool on; 1 val + 2 train steps each, rollouts/rewards inspected per episode):
  1. `@rllm.rollout` + NullEnv + host evaluator (math RL) — rewards 1.0, real token-level updates.
  2. BashHarness + Daytona + host evaluator — secrets baked via Dockerfile read back in observations (proves the snapshot env); one genuine agent hallucination correctly scored 0.0; warm-prefetched rollouts at `setup=0s`.
  3. mini-swe `BaseCliHarness` + Daytona + sandbox-shell verifier — content-addressed snapshot reuse across dataset dirs, warm queue active, in-sandbox `test.sh` grading at 1–2 s/rollout.

A reviewer pass tightened the public API before commit: `Fixed`/`FromTask` → `FixedEvaluation`/`FromTaskEvaluation`, `EnvScan` → `EnvRequirements`, `llm_from_env` → `llm_inside_env`, the no-consumer warning deduped per task id, and history-narrating comments removed.

### DIFF
--- a/rllm/cli/eval.py
+++ b/rllm/cli/eval.py
@@ -38,7 +38,7 @@ def _dict_rows_to_tasks(rows: list[dict]) -> list[Task]:
     ``task.toml`` + ``environment/Dockerfile`` metadata (workdir, verifier
     timeout, per-task image) so per-task verifier resolution and an rllm-native
     harness get the right sandbox. Other rows get ``dataset_dir=Path(".")`` and
-    rely on ``evaluator_override``.
+    rely on the fixed-evaluator policy.
     """
     from pathlib import Path
 
@@ -474,7 +474,7 @@ def _run_eval(
             agent_name=agent_name,
             dataset_name=getattr(dataset, "name", benchmark) or benchmark,
             on_episode_complete=on_episode_complete,
-            evaluator_override=evaluator,
+            evaluator=evaluator,
             sampling_params=(sampling_config.as_dict() if sampling_config is not None else None),
         )
     )
--- a/rllm/cli/train.py
+++ b/rllm/cli/train.py
@@ -213,26 +213,27 @@ def _run_train(
         except (KeyError, ImportError, AttributeError, TypeError) as e:
             fail(f"Cannot load agent '{agent_name}': {e}")
 
-        # Evaluator: --evaluator > train dataset.toml [verifier]; one host-side
-        # evaluator scores both train and val. Sandbox-shell verifiers aren't
-        # supported here — use --evaluator to override.
+        # Evaluator: --evaluator > train dataset.toml [verifier]; a host-side
+        # evaluator scores both train and val. Env-style verifiers
+        # (sandbox-shell / python-hybrid) resolve per task inside the sandbox
+        # via SandboxTaskHooks, so leave ``evaluator`` unset for those.
         if evaluator_name is not None:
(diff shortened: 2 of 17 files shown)


listar2000:
@jeffreysijuntan I will have another follow-up PRs (with this one) targeting some refactoring over agent_flow. This would involve some non-trivial redesign that might benefit from your insights. Will also send out a design doc soon.

jeffreysijuntan:
@listar2000 sounds good

PR #667, opened 2026-06-19
fix: group trajectories by task_id for correct GRPO advantage computation

### LINKED ISSUE #605: AgentWorkflowPPOTrainer: GRPO groups by trajectory.uid instead of prompt/agent key
### Question

I'm not sure whether the current behavior is intentional or I'm misunderstanding the design. Reporting what I observed vs. what I expected — would appreciate maintainer confirmation.

### Expected behavior (as I understand it)

| Config | GRPO grouping `uid` | Are 4 rollouts grouped as one trajectory-level group? |
|--------|---------------------|--------------------------------------------------------|
| `stepwise_advantage.enable=False` | Prompt-level UUID (`task_ids`) | **Yes** |
| `mode=per_step` | `{prompt_uuid}_step{k}` | **No** (group by step index within the same prompt) |
| `mode=broadcast` | Prompt-level UUID; compute advantage on last step only, then broadcast to earlier steps | **Yes** (trajectory-level comparison, then copy to each step) |

### Actual behavior (current code)

Before `compute_advantage`, all three configs do:

```python
batch.non_tensor_batch["uid"] = batch.non_tensor_batch["step_ids"]
```

`step_ids` comes from `trajectory.uid` in `rllm/experimental/verl/transform.py` (a unique UUID per trajectory), which does **not** match the table above:

| Config | Current GRPO `uid` | Are 4 rollouts grouped together? |
|--------|-------------------|----------------------------------|
| `enable=False` | `trajectory.uid` (4 distinct values) | **No** (group size 1 → advantage ≈ raw reward) |
| `mode=per_step` | Still `trajectory.uid` | **No** (neither prompt-level nor `{uuid}_step{k}`) |
| `mode=broadcast` | Still `trajectory.uid` | **No**; also, transform emits all rows with `is_last=True`, so the broadcast path appears largely ineffective |

`stepwise_advantage` currently only changes the reward source (`step_rewards` vs `traj_rewards`) and the broadcast flow — it does **not** change the GRPO grouping key.

### Possible impact

If prompt-level GRPO is intended, multiple rollouts of the same prompt may not be compared within a group. The unified trainer path uses `TrajectoryGroup` for advantage computation and may not be affected.

### Suggested fix (if this is a bug)

- `enable=False` / `mode=broadcast`: use `task_ids` (single agent) or `trajectory_ids` (`{task_id}_{name}`, multi-agent) as GRPO `uid`
- `mode=per_step`: use `{task_id}_step{k}` (or whatever step index format transform uses)
- `broadcast`: may also require transform to emit rows with `is_last_step=False` for intermediate steps

### Minimal repro

1 prompt, `rollout.n=4`, rewards `[1, 0, 1, 0]`:

- **Current**: advantages ≈ `[1, 0, 1, 0]`
- **Expected (if `enable=False`)**: group mean = 0.5 → advantages ≈ `[+0.87, -0.87, +0.87, -0.87]` with `norm_adv_by_std_in_grpo=true`

**Files**: `rllm/trainer/verl/agent_workflow_trainer.py`, `agent_workflow_trainer_fireworks.py`, `rllm/experimental/verl/transform.py`

### Relevant Code / Config

```python

```

### Environment

_No response_

Jackie2049:
## Summary

Fixes #605 — group trajectories by `task_id` instead of `f"{task_id}:{trajectory.name}"` for correct GRPO advantage computation.

## Problem

The current grouping key `f"{task_id}:{trajectory.name}"` causes GRPO groups to have size = 1 when different agents/flows have different trajectory names. With group size = 1:
- `std(group_rewards) = 0` 
- GRPO advantage degrades to raw reward (equivalent to REINFORCE)
- No variance reduction, making GRPO fundamentally broken

## Root Cause

In `rllm/trainer/algorithms/transform.py:127`, trajectories are grouped by `f"{task_id}:{trajectory.name}"`. `trajectory.name` defaults to `_DEFAULT_TRAJ_NAME = "default_traj_name"` or is set to the agent's name. Different agents have different names, so responses to the same prompt end up in separate groups.

## Fix

Change the grouping key from `f"{task_id}:{trajectory.name}"` to `task_id`:

```python
# Before (broken):
trajectories_by_name[f"{task_id}:{trajectory.name}"].append(trajectory)

# After (correct):
trajectories_by_name[task_id].append(trajectory)
```

Same fix applied to metadata grouping on line 128.

## Cross-Framework Evidence

**verl** groups by prompt/task_id — all responses to the same prompt are in one batch group, regardless of which agent produced them.

**DeepSpeed** and **Megatron** also group by prompt for GRPO — the trajectory/agent name is irrelevant to advantage computation.

## Impact

This bug means ALL rLLM GRPO training produces incorrect advantages with zero variance reduction. Combined with #663 (Step.output was None → all rewards = 0.0, now fixed), pre-#663 training was doubly broken.

### DIFF
--- a/rllm/experimental/common/transform.py
+++ b/rllm/experimental/common/transform.py
@@ -122,8 +122,8 @@ def _build_trajectory_groups(episodes: list[Episode], compact_filtering_config: 
         for trajectory in episode.trajectories:
             if len(trajectory.steps) == 0:
                 continue
-            trajectories_by_name[f"{task_id}:{trajectory.name}"].append(trajectory)
-            metadata_by_name[f"{task_id}:{trajectory.name}"].append(
+            trajectories_by_name[task_id].append(trajectory)
+            metadata_by_name[task_id].append(
                 {
                     "episode_id": episode.id,
                     "termination_reason": episode.termination_reason,
--- a/rllm/trainer/algorithms/transform.py
+++ b/rllm/trainer/algorithms/transform.py
@@ -1,0 +1,253 @@
+"""
+Episode to TrajectoryGroup transformation pipeline.
+
+This module provides functions to transform raw Episodes into TrajectoryGroups
+for advantage computation in multi-agent/multi-trajectory workflows.
+
+The pipeline handles:
+1. Trajectory name imputation (for unnamed trajectories)
+2. Reward validation and propagation
+3. TrajectoryGroup construction with configurable grouping strategies
+"""
+
+import logging
+from collections import defaultdict
+from collections.abc import Callable
+
+import numpy as np
+
+from rllm.trainer.algorithms.config import CompactFilteringConfig, TransformConfig
+from rllm.types import Episode, Trajectory, TrajectoryGroup
+from rllm.workflows.workflow import TerminationReason
+
+logger = logging.getLogger(__name__)
+LOG_N_WARNINGS = 1
+
+
+def _impute_trajectory_names(
+    episodes: list[Episode],
+    config: TransformConfig,
+) -> list[str]:
+    """
+    Impute missing trajectory names using position-based naming. This imputation is in-place.
+
+    For each episode, if a trajectory lacks a proper name, rename it to
+    '{prefix}_{position}' (e.g., 'agent_0', 'agent_1').
+
+    Args:
+        episodes: List of episodes to process
+        config: Transform configuration
+
+    Returns:
+        List of warning messages
+    """
+    warnings = []
+    for episode in episodes:
+        # Track position-based names needed
+        new_trajs = []
+        for traj_idx, trajectory in enumerate(episode.trajectories):
+            if not trajectory.name or trajectory.name == config.default_traj_name:  # is unnamed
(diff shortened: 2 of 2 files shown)


Jackie2049:
Note: The `rllm/experimental/common/transform.py` path in the first commit no longer exists on the current main branch (the file was moved to `rllm/trainer/algorithms/transform.py`). The actual fix is in the second commit, which modifies the current file path.

The core change is identical in both files:
- Line 127: `trajectories_by_name[f"{task_id}:{trajectory.name}"]` → `trajectories_by_name[task_id]`
- Line 128: `metadata_by_name[f"{task_id}:{trajectory.name}"]` → `metadata_by_name[task_id]`

I'll create a clean PR with just the 2-line change on the current file path if needed.

jeffreysijuntan:
Can you elaborate on a case where you have different agents but wants them to group together? The trajectory.name is to enable multi-agent trajectories (e.g. solver judge) for example to be grouped based on the agent, so that solver trajectories and judge trajectories are grouped separately.

PR #668, opened 2026-06-19
feat(algorithm): ECHO env-observation loss across verl/tinker/fireworks

jeffreysijuntan:
## What

Adds **ECHO** ([arXiv:2605.24517](https://arxiv.org/abs/2605.24517), *Terminal Agents Learn World Models for Free*) as a selectable algorithm in the unified trainer, working across **all three backends** (verl, tinker, fireworks).

ECHO augments GRPO with an auxiliary cross-entropy loss on **environment-observation tokens** — the tool/terminal output the policy conditions on but that GRPO masks out of its loss. Advantages are identical to GRPO; the only change is the extra loss term. For a terminal/SWE agent this turns every rollout (including the many that fail) into dense supervision, at no extra rollout cost.

## How to use

Flip GRPO → ECHO with one override on any backend:

```bash
bash cookbooks/swe-rl/train_tinker.sh rllm.algorithm.adv_estimator=echo
bash cookbooks/swe-rl/train_verl.sh   algorithm.adv_estimator=echo
```

`adv_estimator=echo` defaults λ to the paper's **0.05**. Override with `rllm.algorithm.env_loss_coef=<λ>` (productive range 0.01–0.05); `0.0` reproduces plain GRPO.

## Design

ECHO is **not** a new advantage estimator (advantages == GRPO) — it's an auxiliary loss term. So the design is a shared backend-agnostic core plus a thin per-backend loss adapter.

**Shared core**
- `rLLMAdvantageEstimator.ECHO` + `env_loss_coef` on `AlgorithmConfig`; `None` resolves to `0.05` for echo else `0.0` (a sentinel distinguishes "unset" from explicit `0`).
- `echo` registered as a GRPO-advantage alias in the estimator registry.

**Per-backend env-CE term** (the only genuinely backend-specific part — each computes its loss differently):
- **verl** — added in `CustomPPOLoss`, reusing the forward pass's per-token log-probs and GRPO's own `loss_agg_mode`/global-batch normalization. **Single pass, exact, free** (as in the paper). Observation tokens = `(responses != pad) & ~response_mask`.
- **tinker** — a gradient-accumulated `cross_entropy` pass (`weights=λ` on obs tokens), returned separately so the policy-pass logprobs stay aligned. Clean because tinker's `optim_step` applies no extra normalization.
- **fireworks** — same second pass, but normalization is folded into the weights/advantages and `GradAccNormalization` is forced to `NONE`, because its counting normalization spans both passes and would otherwise rescale the policy gradient.

## Fidelity note

On verl the env term is free and exact (one shared forward pass). On tinker/fireworks — managed services with fixed server-side loss kernels — it's a second gradient-accumulated `cross_entropy` pass: **no extra rollouts**, but one extra backward, and λ may need light per-backend retuning since loss normalization differs across services. Documented in code + the cookbook README.

## Safety / compatibility

`env_loss_coef` defaults to `0.0` everywhere, so existing GRPO runs are byte-for-byte unchanged.

## Tests

- 5 ECHO cases added to `tests/unified_trainer/test_algorithm_config.py` (config resolution); full file 8/8 pass.
- Numerically verified each backend's env term equals the paper's `λ·L_env = λ·mean_seq((1/|O|)Σ_obs −log p)`.
- All edited files pass `ruff`. (The pre-existing `test_verl_policy_loss.py` failures are unrelated — a stale `patch_verl_actor_for_loss_override` reference + `verl` not installed in CI's base env.)

### DIFF
--- a/cookbooks/swe-rl/README.md
+++ b/cookbooks/swe-rl/README.md
@@ -104,6 +104,37 @@ shape but no 3.5-4B; swap `model.name` / `model.tokenizer_model` /
 model/shape catalog). The synchronous (on-policy) variant is
 `train_fireworks_sync.sh`.
 
+### ECHO (train on environment feedback)
+
+[ECHO](https://arxiv.org/abs/2605.24517) adds a cross-entropy loss on the
+environment-observation tokens (tool/terminal output) that the policy already
+conditions on but GRPO never trains. For a terminal/SWE agent this turns every
+rollout — including the many that fail — into dense supervision, at no extra
+rollout or forward-pass cost. It uses GRPO's advantages unchanged; the only
+difference is the extra loss term.
+
+Flip GRPO → ECHO with one override on any backend (verl / tinker / fireworks):
+
+```bash
+# tinker (async or sync), verl, or fireworks — same switch:
+bash cookbooks/swe-rl/train_tinker.sh    rllm.algorithm.adv_estimator=echo
+bash cookbooks/swe-rl/train_verl.sh      algorithm.adv_estimator=echo
+bash cookbooks/swe-rl/train_fireworks.sh rllm.algorithm.adv_estimator=echo
+```
+
+`adv_estimator=echo` defaults the env-loss weight λ to the paper's 0.05. Tune it
+explicitly with `rllm.algorithm.env_loss_coef=<λ>` (productive range 0.01–0.05;
+`0.0` reproduces plain GRPO). It is implemented as an `env_prediction`
+[auxiliary loss](../../design/auxiliary-losses.md); watch
+`actor/aux_env_prediction_loss` (verl) / `train/aux_*` (tinker, fireworks) to
+confirm the environment-prediction loss is falling.
+
+> On verl the env term shares GRPO's single forward pass (free, exact). On
+> tinker/fireworks (managed training services with fixed server-side loss
+> kernels) it is a second, gradient-accumulated `cross_entropy` pass over the
+> same rollouts — no extra rollouts, but one extra backward. λ may need
+> per-backend retuning since loss normalization differs across services.
+
 ## Evaluation (no training)
 
 ```bash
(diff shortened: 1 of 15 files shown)


jeffreysijuntan:
### Update: ECHO generalized into an `AuxiliaryLoss` framework

Pushed two follow-up commits so the PR now contains both the concrete algorithm and the abstraction it motivated:

1. **`docs(design)`** — `design/auxiliary-losses.md`, the RFC for a general token-level auxiliary-loss framework (with ECHO and SDAR as reference clients).
2. **`refactor(algorithm)`** — prototypes migration steps 1–3 of that doc: ECHO is now the first client of a registry-driven `AuxiliaryLoss` abstraction instead of bespoke per-backend code.

**What changed**
- `rllm/trainer/algorithms/aux_loss.py` — `AuxiliaryLoss` spec + `register_aux_loss` registry + `EnvPredictionLoss` + `build_aux_losses()` (with `env_loss_coef` / `adv_estimator=echo` back-compat sugar). New `algorithm.aux_losses` config list.
- Backend executors replace the bespoke ECHO code: verl `CustomPPOLoss._apply_aux_losses` (single pass), tinker `_get_aux_loss_futures`, fireworks `_build_aux_datums` + the fold/`NONE` logic — all driven by `build_aux_losses()`. Shared managed datum builder in `rllm/trainer/tinker/aux_loss.py`.

**Behavior:** identical for ECHO — `adv_estimator=echo` / `env_loss_coef` still work and resolve to an `env_prediction` aux loss. Default-off unchanged.

**Adding a new aux loss is now one registered class + config, no per-backend surgery.** SDAR's dynamic-weight + teacher-forward extensions (the `requires` field) are the next milestone and currently raise `NotImplementedError`.

Tests: `tests/unified_trainer/test_aux_loss.py` (registry, config/ECHO-sugar resolution, managed datum builder, executor math incl. empty-observation safety) — 24 passing with the config tests; ruff clean.

> Reviewers: this expands the PR beyond the original ECHO scope. Happy to split the refactor into its own PR stacked on the ECHO commit if preferred.

PR #679, opened 2026-06-21
fix(eval): terminal-bench reliability — sandbox lifetime, retry, live metrics, prompt-cache affinity

jeffreysijuntan:
## Problem

Terminal-bench (`terminal-bench@2.0`) eval on Modal scored **~30–33%** vs the official Terminus-2 **~40%**, with flaky `Command failed` warnings. The gap was **infra, not model quality**.

Analysis of three complete 89-task runs: 29/27/29 pass, and **5/6/13 tasks per run auto-zeroed** as 0-step `ERROR` episodes — a *different* set each run (flaky), concentrated in high-`agent_timeout` tasks.

**Root cause:** the Modal sandbox lifetime was derived from a *global* env var (a flat 4800s), while the agent runs for the *per-task* `agent_timeout` (600–12000s) and the verifier then runs in the **same box**. 14 of 89 tasks have `agent+verifier > 4800s`, so the box was reaped before the verifier's exec → *"Sandbox already shut down"* → the whole episode became a bare 0-step `ERROR` (all agent work discarded). With `retry_limit=1` (no retry), each was a permanent zero.

## Changes

- **`eval/_resolution.py`** — size the Modal sandbox per-task: `max(RLLM_MODAL_SANDBOX_TIMEOUT_S, agent_timeout + verifier_timeout + install + slack)`. The box always outlives the agent + verifier; an explicit override only *raises* the ceiling.
- **`eval/runner.py`** — eval `retry_limit` 1 → 2, so a transient infra failure gets one retry instead of a permanent zero (matches the official Harbor harness). Only the few errored tasks re-run.
- **`engine/agentflow_engine.py`** — live running accuracy/reward on the progress bar (`acc N/done=% reward=…`), so the score is visible *as* the eval runs.
- **`sandbox/protocol.py` + `modal_backend.py`** — classify an exec killed at its own `timeout` as a new `SandboxCommandTimeout` (a `RuntimeError` subclass) instead of a generic `Command failed (exit -1)`. **`cli_harness.py`** catches it and logs at INFO (*"reached its time budget"*) — an agent spending its full per-task budget is expected (the verifier still scores partial state), not a failure.
- **`rllm-model-gateway/proxy.py`** — inject `x-session-affinity` / `x-multi-turn-session-id` from the rollout session id on the **HTTP forwarding path**, so multi-turn eval trajectories reuse Fireworks' per-replica prompt cache. Previously only the in-process training path set affinity; eval (the HTTP path) sent none, so 50–100-turn terminus2 rollouts re-prefilled the growing prompt on a fresh replica every turn.

## Validation

- All harness/eval/gateway unit tests pass (`tests/harnesses`, `tests/eval`, gateway proxy logic). Pre-existing failures are unrelated (`verl`/`tinker`/`pytest_asyncio` not installed in this env).
- Per-task lifetime sizing verified: short 900s tasks shrink 4800→3000s; sam-cell-seg → 15600s; build-pov-ray → 25200s. A stale override only raises the ceiling.
- Affinity-header injection verified: headers present with a session id, absent without, client-set values preserved, hop-by-hop still stripped.

## Notes / scope

- `modal_backend.py` also carries a **pre-existing** `_CreateRateLimiter` (Modal create-rate pacing) that shares the new `import time`; it's bundled here only because it lives in the same file and is also eval-reliability related.
- Excludes other in-progress working-tree edits on `terminal-rl` (the `train_*.sh` config tuning and `fireworks_engine.py`).
- Follow-up worth checking: Fireworks prompt caching also needs a **stable prompt prefix** — confirm the terminus2 system prompt keeps dynamic content (timestamps) at the end.

### DIFF
--- a/rllm-model-gateway/src/rllm_model_gateway/proxy.py
+++ b/rllm-model-gateway/src/rllm_model_gateway/proxy.py
@@ -209,7 +209,7 @@ class ReverseProxy:
             # HTTP proxy path
             worker = self.router.route(session_id)
             url = self._build_url(worker.api_url, request.url.path, str(request.url.query))
-            headers = self._forward_headers(request)
+            headers = self._forward_headers(request, session_id)
             try:
                 resp = await self._send_with_retry(
                     method=request.method,
@@ -331,7 +331,7 @@ class ReverseProxy:
         else:
             worker = self.router.route(session_id)
             url = self._build_url(worker.api_url, "/v1/completions", "")
-            headers = self._forward_headers(request)
+            headers = self._forward_headers(request, session_id)
             raw_body = json.dumps(completions_body).encode()
             try:
                 resp = await self._send_with_retry(
@@ -401,7 +401,7 @@ class ReverseProxy:
 
         worker = self.router.route(session_id)
         url = self._build_url(worker.api_url, "/v1/completions", "")
-        headers = self._forward_headers(request)
+        headers = self._forward_headers(request, session_id)
         raw_body = json.dumps(completions_body).encode()
 
         assert self._http is not None
@@ -634,7 +634,7 @@ class ReverseProxy:
 
         worker = self.router.route(session_id)
         url = self._build_url(worker.api_url, request.url.path, str(request.url.query))
-        headers = self._forward_headers(request)
+        headers = self._forward_headers(request, session_id)
 
         assert self._http is not None
         upstream = self._http.stream(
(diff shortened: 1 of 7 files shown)


jeffreysijuntan:
Rebased onto `terminal-rl` after #681 merged. #681 and this PR both added a live postfix to the eval progress bar, which was the only conflict (`agentflow_engine.py`).

Resolved by keeping this PR's version, which **supersedes #681**: #681 showed only `avg_reward=…`, whereas this shows **`acc N/done=X% reward=Y`** — the running *accuracy* (the original ask) plus mean reward. It's a superset, so #681's postfix is dropped in favor of it. All other 6 files are disjoint from #681.

PR #696, opened 2026-06-26
feat(gateway,engine): unify cumulative-mode rendering on one renderer (PR A)

jeffreysijuntan:
## What & why

In cumulative token mode the trajectory was rendered by **two different tokenizers**: turn 0 by the engine's `ChatTemplateParser` (HF `apply_chat_template`) and turns 2+ by the gateway renderer's `bridge_to_next_turn`. The seam between them only *coincidentally* agreed per model (Qwen worked) and crashed engine construction for models whose Jinja template the parser's equivalence self-check can't render (**GLM**: `'str object' has no attribute 'items'`).

This is **PR A** of the renderer-unification plan (PR 0 = `feat/native-renderers`, the base of this PR; PRs B–C = repoint the non-cumulative engine path at `resolve()` and retire `ChatTemplateParser`). It makes **one renderer own the entire trajectory** and reduces the engine to a pure token-in/token-out sampler in cumulative mode.

## Changes

**Gateway (backend-agnostic — fixes Fireworks, Tinker, and vLLM at once):**
- `token_accumulator.py`: `build_first_prompt()` renders turn 0 via `renderer.render_ids(msgs, tools, add_generation_prompt=True)` — the turn-0 analog of `build_next_prompt`.
- `proxy.py`: `handle()` intercepts the **first** turn and sends it token-in through the existing cumulative path (falls back to the chat path if the renderer can't render it — which then seeds the accumulator as before). Completion parsing already flows through `renderer.parse_response` (PR 0).

**Engine (FireworksEngine; `TinkerEngine` untouched):**
- Builds **no** `ChatTemplateParser` in cumulative mode → kills the GLM `get_parser` crash at the source.
- `assemble_model_output` gains a raw-decode branch (gated on `raw_token_mode`, default off → all non-cumulative paths unchanged).
- `get_model_response` raises an actionable error if reached in cumulative mode (the gateway owns rendering).
- Flag plumbed from `cfg.rllm.gateway.cumulative_token_mode`.

## Why the facade makes this correct
The prefix that `bridge_to_next_turn` extends is now the renderer's **own** turn-0 output, so the trajectory is self-consistent by construction for **every** model the registry resolves — not just Qwen.

## Tests (53 passing)
- New: `test_turn0_uses_completions_with_token_ids`, `test_turn0_prompt_equals_render_ids`, `test_turn0_returns_parsed_message`, and **`test_single_renderer_trajectory_no_seam`** (turn-1 extends turn-0's `render_ids` output byte-for-byte).
- New `tests/test_fireworks_engine_cumulative.py`: cumulative construct skips `ChatTemplateParser` (verified with a tokenizer that raises on any access), raw-assemble decodes without parsing, chat path rejected.
- All 40 pre-existing cumulative/accumulator tests + renderer/tinker-adapter/injection suites pass unchanged (the existing `_MockRenderer` has no `render_ids`, so they now exercise the documented fallback).

## Scope / follow-ups
- **Backend-agnostic** via the gateway. Only `FireworksEngine` needed the pure-sampler change; `TinkerEngine` symmetry is a clean follow-up.
- Streaming turn-0 emits raw-text deltas (same as streaming turn-2+ today; the non-streaming training path parses). Pre-existing, noted not fixed.
- **Before merge:** (1) Qwen byte-parity gate — confirm `render_ids`-turn-0 yields the same training rows as the old chat-path turn-0 (the one behavior change for already-working models); (2) GLM end-to-end smoke via `train_fireworks_glm5p2.sh`.

### DIFF
--- a/rllm-model-gateway/src/rllm_model_gateway/proxy.py
+++ b/rllm-model-gateway/src/rllm_model_gateway/proxy.py
@@ -264,6 +264,24 @@ class ReverseProxy:
                             f"{[type(m.get('content')).__name__ for m in new_messages]})"
                         )
                     acc.reset(reason)
+            else:
+                # First turn (turn 0): render the initial messages from scratch via
+                # the renderer and route through the *same* token-in path as later
+                # turns, so the ENTIRE trajectory is tokenized by one renderer — no
+                # chat-template-vs-renderer seam at the turn-0/turn-1 boundary. Falls
+                # back to the chat path if the renderer can't render turn 0 (the chat
+                # path then seeds the accumulator via its own turn-0 ingest below).
+                messages = request_body.get("messages", [])
+                token_ids = acc.build_first_prompt(messages, tools=request_body.get("tools")) if messages else None
+                if token_ids is not None:
+                    return await self._handle_cumulative_turn(
+                        request,
+                        request_body,
+                        session_id,
+                        acc,
+                        token_ids,
+                        originally_requested_logprobs,
+                    )
 
         if is_stream:
             return await self._handle_streaming(request, body, request_body, session_id, originally_requested_logprobs)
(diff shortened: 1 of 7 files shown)


jeffreysijuntan:
Superseded by #699 (same cumulative-mode unification, now stacked on the slim #698 renderer-layer PR instead of the heavy feat/native-renderers branch).

PR #697, opened 2026-06-26
feat(renderers): native renderer layer + unified cumulative-mode rendering

jeffreysijuntan:
Focused replacement for the `feat/native-renderers` PR (which had accumulated unrelated metrics / sandbox / harness / aux-loss / verl-0.8 / cookbook changes) **and** PR #696. This is just the renderer work, rebased clean on `main`.

### 1. Native renderer layer (`rllm/renderers/`)
One token-level `Renderer` protocol — `render_ids` / `parse_response` / `get_stop_token_ids` / `bridge_to_next_turn` — over two ecosystems:
- **prime-rl** (`_prime.py`): passthrough (already token-level, has a real bridge).
- **tinker-cookbook / Fireworks** (`_tinker.py`): adapter that lifts the message-level renderer to the token-level protocol, **including a generic cross-turn bridge** (built from the renderer's own primitives, validated byte-for-byte vs prime-rl's for Qwen3/3.5).
- `registry.resolve()` routes each model **RL-first**: prime-rl exact-match → tinker/FW adapter → DefaultRenderer. The gateway builds/injects the renderer via `resolve()`, so models prime-rl lacks (DeepSeek-V4, …) get a bridge while the gateway package stays free of tinker/Fireworks deps.

### 2. Cumulative-mode unification — one renderer owns the whole trajectory
Previously the trajectory was tokenized by **two** renderers: turn 0 by the engine's `ChatTemplateParser` (HF `apply_chat_template`), turns 2+ by the gateway renderer's bridge. That seam only coincidentally agreed per model (Qwen) and crashed construction for models the parser's self-check can't render (**GLM**).

- The gateway now renders **turn 0** via `renderer.render_ids` and bridges turns 2+ — both from the same renderer — and parses every completion via `renderer.parse_response`.
- `FireworksEngine` is a pure token-in/token-out sampler in cumulative mode (builds **no** `ChatTemplateParser`), fixing the GLM `get_parser` crash at its source.

## Scope
24 files: `rllm/renderers/*`, the gateway/engine cumulative wiring (`proxy.py`, `server.py`, `token_accumulator.py`, `manager.py`, `tinker_adapter.py`, `fireworks_engine.py`, `tinker_engine.py`, `fireworks_backend.py`), tests, and `RENDERER_MERGE_PLAN.md`. **Deliberately excluded** from the old branch: per-task metrics, sandbox download, harness logging, ECHO/aux-loss, verl-0.8 backend changes, and the cookbook scripts. Main's `manager.py` `server_addresses` fix is preserved.

## Tests / lint
73 passing (renderer parity + bridge + DeepSeek-V4; turn-0-via-renderer + single-renderer **no-seam**; FireworksEngine cumulative construct / raw-assemble / rejects-chat-path). `ruff check` + `ruff format` both clean (fixed the formatting the original branch left unrun).

## Before merge
- Qwen byte-parity gate: confirm `render_ids`-turn-0 yields the same training rows as the old chat-path turn-0 (the one behavior change for already-working models).
- GLM end-to-end smoke via `train_fireworks_glm5p2.sh`.

### DIFF
--- a/RENDERER_MERGE_PLAN.md
+++ b/RENDERER_MERGE_PLAN.md
@@ -1,0 +1,328 @@
+# Renderer Merge Plan — unify prime-rl + Fireworks renderers for rLLM training
+
+**Goal:** support more models for rLLM **RL-rollout** training by combining the
+model coverage of two renderer ecosystems behind a single rLLM-side interface,
+preferring the backend that gives the strongest multi-turn token-identity
+guarantee.
+
+**Status:** native renderer package **implemented** (`rllm/renderers/`, tests in
+`tests/test_renderers.py`, 13 passing). Engine/gateway repointing (§6 PR 4–5) is
+**not yet done** — the new layer is additive and not yet wired into the rollout
+engines or the gateway.
+
+### Implemented (this change)
+- `rllm/renderers/` — native token-level `Renderer` protocol + `RenderedTokens` /
+  `ParsedResponse` / `ToolSpec` (reuses `rllm.tools.tool_base.ToolCall`).
+- `resolve()` / `select_backend()` registry with RL-first precedence
+  (prime-rl exact-match → tinker/Fireworks adapter → DefaultRenderer), plus
+  `renderer_family` (prime) and `renderer_name` (tinker) overrides.
+- prime-rl backend wrapper (`_prime.py`) and tinker/Fireworks adapter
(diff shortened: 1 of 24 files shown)


jeffreysijuntan:
Superseded by the two-PR split requested by the author: #698 (native renderer layer) + #699 (cumulative-mode unification, stacked on #698). Duplicate-turn-replay dropped.

PR #673, opened 2026-06-20
feat(cli): eval → curate → SFT loop + unified SFT trainer (tinker/fireworks/verl)

jeffreysijuntan:
## Summary

Adds the **eval → curate → SFT** loop to the rLLM CLI: run an eval with `k` samples per task, curate the trajectories by aggregate metrics into an SFT dataset, and fine-tune on it — all from the CLI. Also refactors SFT around a backend-agnostic dispatcher with pluggable backends (**tinker, fireworks, and verl**). The branch now also tracks **verl 0.8.0** (merged from `main`).

```bash
rllm eval math500 --model <m> --attempts 8
rllm dataset from-eval <run_id> --name math500-rft --filter "0 < avg < 1" --select correct
rllm sft math500-rft --backend verl --gpus 8 --epochs 1     # or --backend tinker/fireworks
```

Design doc: `design/sft-distillation.md`.

## What's included

**Curation (`rllm dataset from-eval`)**
- `rllm/eval/curation.py` — load eval run dirs, pool attempts by stable `task_id` across runs, filter tasks, select trajectories (`correct`/`best`/`best-n`/`shortest`/`all`), lazy-load only chosen episodes, emit `{"messages": ...}` rows with provenance, register via `DatasetRegistry`.
- `rllm/eval/filter_dsl.py` — safe per-task boolean DSL over aggregates (`avg`, `best`, `worst`, `solved`, `n`, `n_correct`, and budget-aware `pass@k` via a `name@k` rewrite + AST node-whitelist). `avg@k` is treated as k-invariant.
- `--dry-run`, task-level train/val holdout.

**Unified SFT trainer**
- `rllm/trainer/sft/` — `SFTSpec` (backend-agnostic input), `SFTBackend` ABC (each backend owns its `fit()`), and `AgentSFTTrainer` as the dispatcher (mirrors the RL stack's `AgentTrainer`/launcher seam).
- `TinkerSFTBackend` (migrated loop) and `FireworksSFTBackend(TinkerSFTBackend)` — Fireworks provisions through the **training-shape path** (`init_fireworks_infra("sft", …)` with a `fireworks_infra` doc), shares the tinker-cookbook data pipeline, and runs a synchronous pipelined loop.
[...]

PR #703, opened 2026-06-26
feat(eval): add GAIA benchmark for native rllm eval

Zayed024:
## Summary

Adds GAIA as a native benchmark so `rllm eval gaia` runs it. GAIA (General AI
Assistants) is a headline long-horizon agentic eval reported across current
frontier model releases. Integration follows the existing rows + transform +
reward_fn pattern (cf. hotpotqa/hle) — no sandbox/Docker. Graded by GAIA's
official quasi-exact-match scorer, ported verbatim for leaderboard-comparable
numbers.

## Type of change

- [x] Feature
- [x] Test

## What changed

- `rllm/eval/reward_fns/gaia.py`: `gaia_reward_fn` (`evaluate`) + the official
  `question_scorer` (number / list / string normalization), attributed to the
  GAIA scorer (Apache-2.0).
- `rllm/data/transforms.py`: `gaia_transform` — maps GAIA rows to standard QA
  format; skips file-attachment tasks (text-only subset for now).
- `rllm/eval/evaluator_loader.py`: register `gaia_reward_fn`.
- `rllm/registry/datasets.json`: `gaia` entry (source `gaia-benchmark/GAIA`,
  validation split, `gated: true`, `default_agent: search`).
- `tests/eval/test_gaia_reward.py`: 17 tests — scorer (number/list/string edge
  cases), transform filtering, `evaluate` via `load_evaluator`, registry wiring.

## Validation

- [x] `pytest tests/eval/test_gaia_reward.py -q` → 17 passed
- [x] `ruff check` + `ruff format --check` clean (pinned v0.11.4)
- [ ] Full `rllm eval gaia` run pending (GAIA is HF-gated — needs `HF_TOKEN` —
      and a search-capable agent; evaluator/transform unit-tested in isolation)

## Open questions for reviewers

- Scope: text-only subset first (file-attachment tasks skipped) — want file-tool
  support as a follow-up, or in this PR?
- `default_agent: search` — is that the right harness for GAIA, or prefer `react`?
- τ²-bench is the natural next agentic integration — is it on the roadmap / in flight?

- None

## Related issues / PRs

- None


jeffreysijuntan:
Might need some investigation what harness frontier model use to evaluate on GAIA, and then implement that in rLLM. Can you past screenshots of results from `rllm eval gaia` to confirm the whole thing is working E2E?

Zayed024:
sure i will

Zayed024:
E2E confirmed: rllm eval gaia with gpt-4o → 25% (3/12), 0 errors, agent does search→extract→answer. Fixed a
 bug (one Step per LLM turn) that tripped the trace enricher on parallel tool calls.
Screenshot 2026-07-07 182506

Zayed024:
Pushed a round of hardening after an independent review pass. Summary of what changed:

**Agent robustness (`cookbooks/gaia/agent.py`)**

- Tool execution is now wrapped: a raising tool (missing `TAVILY_API_KEY`, unexpected model-emitted kwargs, network error) turns into an error observation the model can react to, instead of failing the rollout. This matches `ToolCallingMixin`’s behavior in `rllm/harnesses/tool_calling.py`. 
- A failed LLM call no longer appends a `Step`. There is no gateway trace for a call that did not happen, and `enrich_episode_with_traces` requires step↔trace parity, so the previous “graceful” error step actually guaranteed an `EnrichMismatchError`. 
- On `max_turns` exhaustion, the last step is marked `done` (flag flipped, not a new step appended, again for parity); the answer stays empty and is scored as such. 

**Demo removal**

- Removed `cookbooks/gaia/demo_direct.py`. It was based on a diagnosis that turned out to be wrong (I had suspected the eval gateway dropped tools; it does not — the empty responses were due to provider-specific tool-calling support), and the rationale was misleading, so the file is removed rather than corrected.

**Tests**

- Added `tests/eval/conftest.py` (replaces a `sys.path` hack) and expanded coverage for previously untested paths: tool exceptions, parallel `tool_calls` in one turn (the case the one-Step-per-turn fix addresses), malformed tool arguments, `max_turns` exhaustion, LLM error. 
- Extended scorer coverage for semicolon-separated lists, currency/percent list elements, and answers without a `FINAL ANSWER:` prefix. 
- Noted a quirk of the official scorer: list splitting happens before numeric normalization, so a thousands separator inside an element (for example, `\$1,000`) changes the element count and fails the match, which is deliberate for leaderboard comparability. 
- All tests are now green: 31/31 tests pass; `ruff check` and `ruff format` are clean on v0.11.4. 

**Open item**

- Still open: `default_agent: "search"` currently does not resolve to a registered agent. I kept this consistent with the existing `browsecomp` / `hle_search` entries, but if a built-in search agent should be registered (which would make `rllm eval gaia` work out of the box), I am happy to add that here or in a follow-up.

Zayed024:
Apologies for the long silence, this fell off my plate for a while, but I picked it back up and closed out the open item from my Aug 2 comment. Also rebased onto latest `main` (clean; the new eval CLI from #673 doesn't change any of the wiring here).
 
**What changed in this push**
 
- **Built-in `search` agent:** the GAIA agent is promoted to `rllm/harnesses/search.py` (`SearchHarness`) and registered in `registry/agents.json` as `"search"`, so `rllm eval gaia` now works with no `--agent` flag.
- This fixes the dangling `default_agent: "search"` for **all five** datasets that reference it (`gaia`, `browsecomp`, `hle_search`, `seal0`, `widesearch`) — previously `load_agent("search")` raised `KeyError` for every one of them.
- `cookbooks/gaia/agent.py` is now a thin re-export, so `--agent cookbooks.gaia.agent:agent` keeps working unchanged.
- Re: the harness question — `SearchHarness` follows the reference GAIA scaffolds (HF Open Deep Research / Princeton HAL): a search + browse ReAct loop over Tavily, one LLM call per step to keep trace parity. Happy to extend it toward whatever scaffold frontier labs report GAIA numbers with if you'd still prefer that direction.
- +2 tests: `search` resolves via `load_agent`, and every catalog dataset with `default_agent: search` resolves to the harness.
 
**Validation**
 
- `pytest tests/eval/test_gaia_agent.py tests/eval/test_gaia_reward.py -q` → 33 passed
- `ruff check` + `ruff format --check` clean (v0.11.4)
 
Scope note: still the text-only subset (file-attachment tasks skipped). I kept file/code-execution tools out of `SearchHarness` so it stays a clean default for all five search benchmarks — happy to add them for GAIA levels 2–3 as a follow-up.

PR #721, opened 2026-07-02
Reduce memory usage for VLM eval loading

Cozy000000:
## Summary

  Reduce peak memory usage when loading and evaluating large VLM benchmarks. This matters for VLM eval runs because large image payloads were previously encoded eagerly and could be duplicated again for multi-attempt pass@k evaluation.

  ## Type of change

  - [ ] Feature
  - [x] Fix
  - [ ] Docs
  - [ ] Refactor
  - [ ] Example / Project
  - [ ] Infra / CI

  ## What changed

  - Load JSONL rows incrementally while building tasks, avoiding an intermediate full row list.
  - Encode VLM image data URIs lazily when multimodal instructions are consumed instead of during benchmark loading.
  - Add a bounded LRU cache for image data URI encoding to avoid repeated work for reused image paths.
  - Avoid materializing expanded task lists for `--attempts` / pass@k evaluation by using a lightweight repeated-task sequence view.
  - Add regression tests for lazy VLM image encoding and non-copying pass@k task repetition.

  ## Validation

  - [ ] Not run (reason below)

  Validation details:

  ```bash
  conda run -n rllm python -m pytest tests/eval/test_materialize.py tests/tasks/test_loader_memory.py tests/eval/test_runner_memory.py
  conda run -n rllm python -m ruff check rllm/tasks/loader.py rllm/eval/runner.py tests/tasks/test_loader_memory.py tests/eval/test_runner_memory.py
  ```
  Result: 17 passed; ruff passed.

  None

  ## Docs / examples

  - [x] Not needed
  - [ ] Updated docs
  - [ ] Updated examples
  - [ ] Follow-up docs needed

  ## Related issues / PRs

  - Fixes #
  - Related to #
  - Stacked on / depends on #

  ## Screenshots / logs

  Not applicable.

### DIFF
--- a/rllm/eval/runner.py
+++ b/rllm/eval/runner.py
@@ -78,10 +78,8 @@ async def run_dataset(
     from rllm.gateway.manager import EvalGatewayManager
     from rllm.gateway.tunnel import is_local_sandbox_backend
 
-    # pass@k: expand each task into `attempts` adjacent copies. Everything
-    # downstream (warm queue, engine, episode callbacks) sees one entry per
-    # rollout; only the EvalItem aggregation below folds attempts back onto
-    # their task index.
+    # pass@k: expose each task as `attempts` adjacent rollouts while reusing
+    # the same Task objects.
     if attempts > 1:
         tasks = [task for task in tasks for _ in range(attempts)]
 
--- a/rllm/tasks/loader.py
+++ b/rllm/tasks/loader.py
@@ -25,7 +25,9 @@ from __future__ import annotations
 import json
 import logging
 import os
+from collections.abc import Iterator
 from dataclasses import dataclass, field
+from functools import lru_cache
 from pathlib import Path
 
 import tomllib
@@ -34,6 +36,8 @@ from rllm.tasks.dataset_config import DatasetConfig, load_dataset_config
 from rllm.types import Task
 
 logger = logging.getLogger(__name__)
+
+_VLM_IMAGE_DATA_URI_CACHE_SIZE = int(os.environ.get("RLLM_VLM_IMAGE_CACHE_SIZE", "256"))
 
 
 @dataclass
@@ -114,12 +118,11 @@ def _load_data_dataset(
 ) -> BenchmarkResult:
     """Materialised-style dataset: jsonl rows + shared verifier."""
     data_file = _resolve_data_file(path, config)
-    rows = _load_jsonl(data_file)
 
     instruction_field, metadata_fields = _read_dataset_meta(path)
     category = _read_dataset_category(path)
     tasks: list[Task] = []
-    for idx, row in enumerate(rows):
+    for idx, row in enumerate(_iter_jsonl(data_file)):
         text = _render_instruction(path, row, instruction_field)
         if category == "vlm":
             instruction: str | list[dict] = _build_multimodal_instruction(text, row, path)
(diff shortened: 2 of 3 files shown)


jeffreysijuntan (commented):
### Is `_RepeatedTaskSequence` needed? I don't think so.

The lazy-encoding + LRU-cache changes in `loader.py` are the real fix here. But `_RepeatedTaskSequence` solves a non-problem and I'd suggest dropping it.

**It doesn't save meaningful memory.** The line it replaces —

```python
tasks = [task for task in tasks for _ in range(attempts)]
```

— does not copy Task objects or their image payloads. It builds a list of *references* to the same Task objects, i.e. one pointer (8 bytes) per rollout. So the expanded list costs `N × attempts × 8 bytes`. Even at 5,000 VLM tasks × 8 attempts that's ~320 KB of pointers vs. ~40 KB for the underlying list — negligible next to image payloads that are MBs each. The sequence view only avoids that pointer array; the actual Task objects (and their instructions) are shared identically either way (`repeated[i] is tasks[i // attempts]` matches what the list comprehension already did).

**The memory win is entirely from the other change.** `_LazyMultimodalInstruction` + `@lru_cache` on `_image_file_to_data_uri` mean images are encoded only on consumption and deduped across reuse. That's orthogonal to task expansion. The PR description's justification — *"avoid materializing expanded task lists... matters for VLM tasks whose instructions may contain large image payloads"* — doesn't hold: the expanded list never held image payloads, only references to shared Task objects.

**The benefit is partially defeated anyway.** The warm-queue path does `WarmQueue(list(tasks), ...)` (runner.py:129), which re-materializes the whole expanded list. It's only skipped for VLM because there's no sandbox backend — but it shows the view is fragile.

**It's a latent maintenance trap.** A `Sequence` view (slice + negative-index handling, ~25 lines + a dedicated test) silently breaks or re-materializes for any future consumer that assumes a real list — mutation, `.append`/`.copy`, `random.shuffle`, pickling across a process boundary, etc. Today it's only iterated and `len()`'d in `execute_tasks`, so it's *safe* — but it buys nothing for that risk.

**Suggestion:** keep the `loader.py` changes, revert `runner.py:85` to the plain list comprehension, and drop `_RepeatedTaskSequence` + `tests/eval/test_runner_memory.py`. Same behavior, ~50 fewer lines, no fragile Sequence contract.

jeffreysijuntan:
Thanks for your PR. Can you read the above comments generate from Claude and remove the RepeatedTaskSequence as that is not needed for lazy loading?

Cozy000000:
Thanks for the review. Claude is right, I removed `_RepeatedTaskSequence` and the related test as suggested.

  The PR now keeps the focused memory optimization in the loader: JSONL rows are processed incrementally, VLM image data URIs are encoded lazily when multimodal instructions are consumed, and repeated image paths are deduplicated through a bounded LRU cache. The pass@k path now uses the original list-
  comprehension behavior again, since it only duplicates task references and is not the meaningful source of memory growth here.

  Validation:
  - `conda run -n rllm python -m pytest tests/eval/test_materialize.py tests/tasks/test_loader_memory.py tests/eval/test_results.py tests/eval/test_curation.py`
  - `conda run -n rllm python -m ruff check rllm/tasks/loader.py rllm/eval/runner.py tests/tasks/test_loader_memory.py`

PR #754, opened 2026-07-15
feat(sandbox): add Sailboxes as a sandbox backend

jeffreysijuntan:
## Summary

Adds **Sailboxes** ([Sail Research](https://docs.sailresearch.com/sailboxes)) as a pluggable sandbox backend alongside Modal and Daytona. Select it with `sandbox_backend="sailboxes"` (or `--sandbox-backend sailboxes`). Auth via `SAIL_API_KEY` (or `sail auth login`); SDK: `pip install sail`.

New `SailboxesSandbox` implements the `Sandbox` protocol — `exec` / `upload_file` / `upload_dir` / `close` / `is_alive` — plus `set_env`, matching the Modal/Daytona exec conventions (`bash -c`, persistent-env exports, `su` user-switch, `SandboxCommandTimeout`). Wired into `create_sandbox`, the three snapshot dispatchers (`build_snapshot` / `delete_snapshot` / `snapshot_absent`), and `_sandbox_resource_kwargs`.

## Why it fits well

Sailboxes' native **checkpoint / from_checkpoint** maps directly onto rLLM's snapshot triad — arguably a cleaner fit than Modal/Daytona for RL rollouts (their docs describe the exact "set up once, fork/checkpoint per task" pattern):

| rLLM | Sailboxes SDK |
|---|---|
| `exec(cmd, timeout, user)` | `sb.run(["bash","-c",…], timeout=, check=False)` → `ExecResult{stdout,stderr,exit_code,timed_out}` |
| `upload_file` / `upload_dir` | `sb.fs.write(...)` (+ tar→write→extract for dirs) |
| `is_alive` | `Sailbox.get(id).status ∈ {running, sleeping}` |
| `close` | `sb.terminate()` |
| `build_snapshot` | `sb.checkpoint(name, ttl_seconds)` → `checkpoint_id` |
| boot from snapshot | `Sailbox.from_checkpoint(id)` |

## Known limitations (documented in the module docstring)

- **No OCI registry pull.** Sailboxes images build from a fixed Debian base (`pip_install`/`apt_install`/`run_commands`), so a Docker image string falls back to the Debian base + the Dockerfile `RUN` replay. Tasks whose environment lives *in* a prebuilt registry image (SWE-bench / R2E-Gym per-instance images) are **not reproducible** on this backend. Fine for compute/exec-style tasks (countdown, math, pip/apt-installable envs).
- **No checkpoint GET/list/delete API.** The prune-side probes (`_sailboxes_ref_absent`, `delete_sailboxes_snapshot`) are therefore conservative no-ops — they never prune what can't be verified gone; checkpoints self-expire via `ttl_seconds` (default 30d, `RLLM_SAILBOX_CHECKPOINT_TTL_S`).

## Testing

`tests/sandbox/test_sailboxes_backend.py` covers the missing-SDK install hint, the pure exec-command builder (env + `su` user-switch), image classification, and the inert prune/delete probes. Full `tests/sandbox/` suite passes (98) and `ruff check`/`format` are clean. Not yet exercised against a live Sailbox (no `SAIL_API_KEY` in CI); the live path is best-effort verified against the SDK surface (`sail==0.4.2`).

## Base branch

Targets `feat/scaleswe-cookbook` (not `main`) because it depends on the current sandbox abstraction — `SandboxCommandTimeout`, `set_env`, `build_snapshot(prior_ref)`, `_should_replay_dockerfile` — which isn't on `main` yet.

### DIFF
--- a/rllm/eval/_resolution.py
+++ b/rllm/eval/_resolution.py
@@ -335,7 +335,8 @@ def _sandbox_resource_kwargs(task: Task, backend: str) -> dict:
     Harbor task.toml declares ``cpus`` / ``memory_mb`` / ``storage_mb``; without
     these a remote sandbox runs at the backend default (Daytona: 1 GiB), which
     OOM-kills compile-heavy graders (e.g. ``go test ./...``). Modal takes memory
-    in MB; Daytona takes memory/disk in GB. Docker/local ignore the values.
+    in MB; Daytona takes memory/disk in GB; Sailboxes takes memory/disk in GiB.
+    Docker/local ignore the values.
 
     Those per-task values are baked into ``task.toml`` at dataset-build time, so
     an over-provisioned default can only be shrunk by rebuilding the dataset.
@@ -429,6 +430,15 @@ def _sandbox_resource_kwargs(task: Task, backend: str) -> dict:
         # 30-min idle can reap a long task, e.g. during a stalled LLM call that
         # looks idle). Express the shared lifetime floor in minutes, rounded up.
         kw["auto_stop_interval"] = (lifetime_s + 59) // 60
+    elif backend == "sailboxes":
+        # Sailboxes takes memory/disk in GiB; vCPU derives from the size tier, so
+        # ``cpus`` is not mapped. Sailboxes are persistent (no lifetime cap), so
+        # ``lifetime_s`` doesn't translate to a create knob — boot-wait uses the
+        # backend default.
+        if mem_mb:
+            kw["memory_gib"] = max(1, round(mem_mb / 1024))
+        if disk_mb:
+            kw["disk_gib"] = max(1, round(disk_mb / 1024))
     return kw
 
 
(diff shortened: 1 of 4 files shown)


jeffreysijuntan:
Closing for now — not merging yet.

Blocker: Sailboxes (sail==0.4.2) has no registry/OCI/Dockerfile image source. The `ImageSpec` proto's `source` oneof only implements `base` (DEBIAN | DEVBOX) + build steps (apt/pip/run/add_local); a Docker image string falls back to the Debian base. So this backend can't boot the prebuilt per-instance images that SWE-bench/R2E-Gym/scaleswe and fully-built terminal-bench tasks rely on (docker_image set → replay_dockerfile=false).

The proto has a reserved oneof slot (field 6), suggesting a registry source may be planned. Revisit if/when Sail ships one (or via an offline OCI-rootfs import spike).

Branch `feat/sailboxes-backend` kept; work is preserved.

PR #772, opened 2026-07-22
feat(trainer): multi-lineage (DAG) prefix merge for subagent trajectories

jeffreysijuntan:
## Problem

Training runs with the **opencode** harness (and any subagent-spawning CLI harness — claude-code too) show a low `batch/merge_compression_ratio` (`= total_agent_steps / total_emitted_rows`; `1` means no merging, `N` means an N-turn trajectory collapsed to one row).

**Root cause.** opencode's `task`-tool subagent runs inside the same process, which points at one session-scoped gateway URL `/sessions/<uid>/v1`. The gateway derives the session id purely from that path, so **every** LLM call — parent agent *and* subagent — lands in one session / one `Trajectory`, interleaved in time. But a subagent turn is a fresh conversation with a **different system prompt and tool set**, so it is *not* a byte-prefix-extension of the parent. The verl/tinker transforms kept a **single** running segment and reseeded on any non-extension, so `A1, B1, A2, B2, …` never re-merged: once `A1`'s segment was emitted, `A2` could no longer rejoin it → one row per turn → ratio → 1.

## Change (Layer 1 — trainer side)

Replace the single running segment with a **set of open segments** (a forest/DAG of lineages) in both `_process_trajectory` (verl) and `trajectory_to_datums` (tinker):

1. Each step attaches to the **deepest** open segment whose `full_seq` is a byte-prefix of its `prompt_ids`.
2. A step that extends no open segment opens a **new** segment (new lineage root).
3. Emit every open segment as one row.

Parent turns merge among themselves; subagent turns merge among themselves. `A1,B1,A2,B2` → `{A1,A2}` + `{B1,B2}` = 2 rows instead of 4.

**Safety / compatibility**
- Byte-exact prefix match = the same cumulative-extension invariant used today; merging is never lossy.
- A single-lineage trajectory keeps exactly one segment → **byte-identical** to prior behavior. Sequential non-prefix breaks keep the same row count.
- Multi-row-per-trajectory was already supported (`_emit` keys rows by `trajectory.uid` with a broadcast scalar advantage). This only *reduces* row count (`#turns` → `#lineages`), improving per-trajectory weight uniformity.
- Runs on **already-stored** `prompt_ids` — **no gateway change, no rerun** needed.

## Tests
- tinker: `test_interleaved_subagent_lineage_remerges` (parent–subagent–parent → 2 Datums, parent mask intact), `test_two_subagents_between_parent_turns` (parent + 2 subagents → 3).
- verl: `test_interleaved_subagent_lineage_remerges` (→ 2 rows, parent mask `[1,1,0,1,1]`).
- The `verl` module isn't importable in this env (pre-existing: the whole `test_verl_transform.py` can't be collected without it); the merge algorithm was additionally validated with a standalone replica. Two unrelated `test_tinker_transform.py` single-step failures are **pre-existing float32-precision** asserts on `main`, not touched by this PR.

## Follow-up (Layer 2 — proposed, in `design/subagent-dag-merge.md`)
Gateway multi-slot `TokenAccumulator`: route each request to the slot it extends (new slot instead of `reset()` on divergence), keeping generation drift-free across lineages and tagging each trace `parent_trace_id` to **materialize** the DAG — composing with the Phase-1 delta-chain work in `design/gateway-dag-token-storage.md` and implementing its deferred §7 branching case.

### DIFF
--- a/design/subagent-dag-merge.md
+++ b/design/subagent-dag-merge.md
@@ -1,0 +1,142 @@
+# Design: Multi-lineage (DAG) prefix merge for subagent trajectories
+
+- **Status:** Layer 1 implemented (trainer-side multi-segment merger in `rllm/trainer/verl/transform.py` + `rllm/trainer/tinker/transform.py`, with tests). Layer 2 (gateway multi-slot accumulator) proposed.
+- **Related:** `design/gateway-dag-token-storage.md` (delta-chain token storage — this is its deferred §7 "full branching DAG", motivated by subagents); `PR_dev-multiturn-merged-rows.md` (the `merge_compression_ratio` metric); `rllm-model-gateway` `token_accumulator.py` / `proxy.py`; `rllm/harnesses/{opencode,cli_harness}.py`.
+- **Scope:** how a multi-turn rollout whose turns do **not** all share one growing prefix (a subagent runs mid-rollout) is merged into training rows, and (proposed) how the gateway should tag such turns.
+
+---
+
+## Summary
+
+`merge_compression_ratio = total_agent_steps / total_emitted_rows` measures how well a
+trajectory's turns collapse into shared-prefix training rows (`= N` for a clean N-turn
+cumulative rollout → 1 row; `= 1` when nothing merges). Under the **opencode** harness with
+**subagents** the ratio collapses toward 1.
+
+Root cause: opencode's `task`-tool subagent runs **inside the same process**, which points at a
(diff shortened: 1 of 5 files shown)


jeffreysijuntan:
Superseded by a new PR against `terminal-rl` (the branch with the latest gateway implementation), per maintainer direction. Same Layer 1 change; gateway Layer 2 will follow there too.

PR #773, opened 2026-07-22
feat(trainer): multi-lineage (DAG) prefix merge for subagent trajectories

jeffreysijuntan:
## Problem

Training with the **opencode** harness (and any subagent-spawning CLI harness — claude-code too) shows a low `batch/merge_compression_ratio` (`= total_agent_steps / total_emitted_rows`; `1` means no merging).

opencode's `task`-tool subagent runs inside the same process → one session-scoped gateway URL `/sessions/<uid>/v1`, so parent + subagent calls land in one session / one `Trajectory`, interleaved in time. A subagent turn has a **different system prompt/tool set**, so it is not a byte-prefix-extension of the parent. The verl/tinker transforms kept a **single** running segment and reseeded on any non-extension, so `A1,B1,A2,B2,…` never re-merged: once `A1`'s segment was emitted, `A2` could not rejoin it → one row per turn → ratio → 1.

## Change (Layer 1 — trainer side)

Keep a **set of open segments** (a forest/DAG of lineages) in both `_process_trajectory` (verl) and `trajectory_to_datums` (tinker):
1. Each step attaches to the **deepest** open segment whose full sequence is a byte-prefix of its `prompt_ids`.
2. A step that extends none opens a **new** segment (new lineage root).
3. Emit every open segment as one row.

Parent turns merge among themselves; subagent turns merge among themselves.

**Safety / compatibility**
- Byte-exact prefix match = the same cumulative-extension invariant used today; merging is never lossy.
- Single-lineage trajectory keeps exactly one segment → **byte-identical** to prior behavior; sequential non-prefix breaks keep the same row count.
- Multi-row-per-trajectory was already supported (`_emit` keys rows by `trajectory.uid` with a broadcast scalar advantage); this only *reduces* row count.
- Runs on already-stored `prompt_ids` — **no gateway change, no rerun**.

## Tests
- tinker: `test_interleaved_subagent_lineage_remerges` (parent–subagent–parent → 2 Datums, parent mask intact), `test_two_subagents_between_parent_turns` (→ 3).
- verl: `test_interleaved_subagent_lineage_remerges` (→ 2 rows, parent mask `[1,1,0,1,1]`).
- `verl` isn't importable in this env (whole `test_verl_transform.py` uncollectable — pre-existing); merge algorithm validated standalone. Two `test_tinker_transform.py` single-step failures are **pre-existing float32-precision** asserts on `terminal-rl`, not touched here.

## Follow-up (Layer 2, needed for `cumulative_token_mode`)
Under `cumulative_token_mode` the gateway `TokenAccumulator` resets on each subagent switch and re-tokenizes the parent's resume turn, which can drift off the pre-subagent tokens — so Layer 1 alone may leave the parent split at those boundaries. Layer 2 (separate PR, gateway) makes the accumulator multi-slot so the parent lineage stays bridged (drift-free) straight through the subagent, and tags each trace with a lineage/parent pointer. Design: `design/subagent-dag-merge.md`.

### DIFF
--- a/design/subagent-dag-merge.md
+++ b/design/subagent-dag-merge.md
@@ -1,6 +1,6 @@
 # Design: Multi-lineage (DAG) prefix merge for subagent trajectories
 
-- **Status:** Layer 1 implemented (trainer-side multi-segment merger in `rllm/trainer/verl/transform.py` + `rllm/trainer/tinker/transform.py`, with tests). Layer 2 (gateway multi-slot accumulator) proposed.
+- **Status:** Layer 1 implemented (trainer-side multi-segment merger in `rllm/trainer/verl/transform.py` + `rllm/trainer/tinker/transform.py`, with tests) — PR #773. Layer 2 implemented (gateway multi-slot accumulator, `SessionSlots` in `rllm-model-gateway`) — PR #774; required when `cumulative_token_mode` is on.
 - **Related:** `design/gateway-dag-token-storage.md` (delta-chain token storage — this is its deferred §7 "full branching DAG", motivated by subagents); `PR_dev-multiturn-merged-rows.md` (the `merge_compression_ratio` metric); `rllm-model-gateway` `token_accumulator.py` / `proxy.py`; `rllm/harnesses/{opencode,cli_harness}.py`.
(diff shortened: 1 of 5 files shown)


jeffreysijuntan:
Superseded by #775 + #774: the lineage split moves to collection time (a subagent becomes its own trajectory, tagged by the gateway) with a per-rollout advantage dedup, so the transform stays a plain linear merge instead of a multi-slot forest. Same training rows, cleaner separation of concerns.

PR #776, opened 2026-07-22
fix(verl): group GRPO advantage by task, not per-trajectory uid

### LINKED ISSUE #605: AgentWorkflowPPOTrainer: GRPO groups by trajectory.uid instead of prompt/agent key

AmirF194:
## Summary

`AgentWorkflowPPOTrainer` groups rows for GRPO (and other group-relative advantage
estimators) by `batch.non_tensor_batch["uid"]`, but that field was set to `step_ids`,
which is `Trajectory.uid`, a fresh id per trajectory instance. Every `rollout.n` repeat
of the same prompt got a distinct uid, so verl's `compute_grpo_outcome_advantage` grouped
each repeat alone and hit its size-1 branch (`mean=0`, `std=1`), silently dropping the
group baseline whenever `stepwise_advantage.enable=False` (the default) or `mode=broadcast`.

## Type of change

- [x] Fix

## What changed

- `AgentWorkflowPPOTrainer` sets `uid` from the new `task_ids` field instead of `step_ids`.
- `AccumulatedData` gains a `task_ids` list, populated in `_process_trajectory` from the
  `task_id` already threaded through `_process_episode`/`_process_trajectory_group`, and
  carried into the batched `DataProto` in `_batch_tensors_and_build_data_proto`.
- `task_id` is what `interleave_tasks` shares across all `rollout.n` repeats of one task,
  unlike `step_ids`, so rows for the same prompt now land in the same GRPO group.

## Validation

- [x] `pre-commit run --files <changed files>` (ruff, ruff-format): clean
- [x] Targeted tests: `pytest tests/unified_trainer/test_verl_transform.py -k TestTaskIdsForGrpoGrouping`

Validation details:
- Docker differential (`python:3.11-slim`, `verl==0.8.0` from PyPI, `rllm` installed
  `--no-deps` to skip the CUDA-only extras): the new
  `test_task_ids_shared_across_rollout_repeats` fails on `main` (`task_ids` absent from
  the transform output) and passes on this branch.
- `test_grpo_advantage_degenerates_on_step_ids_task_ids_fixes_it` calls the real
  `verl.trainer.ppo.core_algos.compute_grpo_outcome_advantage` (installed from PyPI, not
  vendored) directly: grouping two same-task rollouts (rewards 1.0, 0.0) by `step_ids`
  gives `[1.0, 0.0]` (raw reward, no baseline); grouping the same pair by `task_ids` gives
  `[0.7071, -0.7071]` (mean-subtracted, std-normalized per verl's default
  `norm_adv_by_std_in_grpo=True`).
- Full `tests/unified_trainer/` suite: 43 passed, no regressions (two `test_tinker_*`
  files fail to collect on this environment because the optional `tinker` package isn't
  installed; that is pre-existing and unrelated to this change, and the `test-tinker.yml`
  CI workflow only triggers on paths under `rllm/trainer/tinker/**` / `rllm/engine/rollout/**`,
  neither of which this PR touches).
- Not run: an actual GPU training run through verl. This is a source-level fix verified
  by installing the real `verl==0.8.0` package and calling its own advantage function; I
  don't have a GPU/verl training environment to confirm the end-to-end training loop.

- None. `uid` changes what it groups by, but its shape and consumers are unchanged.

## Docs / examples

- [x] Not needed

## Related issues / PRs

- Fixes #605

### DIFF
--- a/rllm/trainer/verl/agent_workflow_trainer.py
+++ b/rllm/trainer/verl/agent_workflow_trainer.py
@@ -387,10 +387,12 @@ class AgentWorkflowPPOTrainer(RayPPOTrainer):
                             batch = batch.union(values)
 
                     with marked_timer("adv", timing_raw, color="brown"):
-                        # step_ids is safe to always use for advantage computation
-                        # if we're not using computing advantages stepwise (i.e., for cumulative agents or single turn workflows)
-                        # then step_ids == trajectory_ids
-                        batch.non_tensor_batch["uid"] = batch.non_tensor_batch["step_ids"]
+                        # GRPO (and other group-relative estimators) group rows by "uid" and need
+                        # every rollout.n repeat of the same task in the same group. step_ids is
+                        # trajectory.uid, a fresh id per trajectory instance, so it puts every
+                        # repeat in its own group of one; task_ids is the id interleave_tasks()
+                        # shares across a task's repeats and is what grouping needs.
+                        batch.non_tensor_batch["uid"] = batch.non_tensor_batch["task_ids"]
 
                         if self.config.rllm.stepwise_advantage.enable and self.config.rllm.stepwise_advantage.mode == "per_step":
                             batch.batch["token_level_scores"] = batch.batch["step_rewards"]
--- a/rllm/trainer/verl/dataclass.py
+++ b/rllm/trainer/verl/dataclass.py
@@ -42,6 +42,7 @@ class AccumulatedData:
     # ID tracking (parallel to tensor lists)
     trajectory_ids: list[str] = field(default_factory=list)
     step_ids: list[str] = field(default_factory=list)
+    task_ids: list[str] = field(default_factory=list)  # shared across all rollout.n repeats of one task, for GRPO grouping
     episode_ids: list[str] = field(default_factory=list)  # unique identifier for each rollout
 
     # Metadata (parallel to tensor lists)
@@ -74,6 +75,7 @@ class AccumulatedData:
         self,
         step_data: ProcessedStepData,
         trajectory_id: str,
+        task_id: str,
         traj_reward: float,
         step_num: int,
         is_last: bool,
@@ -94,6 +96,7 @@ class AccumulatedData:
             self.advantages.append(step_data.advantage)
 
         self.trajectory_ids.append(trajectory_id)
+        self.task_ids.append(task_id)
         self.step_nums.append(step_num)
         self.is_last_step.append(is_last)
         self.multi_modal_inputs.append(step_data.multi_modal_inputs)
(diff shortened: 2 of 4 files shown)


AmirF194:
No rush, just checking in after about a week. Still clean and mergeable on my end (the two held checks are awaiting approval to run, not failures), happy to rebase or add anything else useful for review.

AmirF194:
Closing to keep the queue clean; happy to reopen if there's interest.

PR #777, opened 2026-07-23
fix(tasks): drop phantom 600s agent_timeout default in the task loader

jeffreysijuntan:
## Problem

When a task's \`task.toml\` doesn't declare \`[agent].timeout_sec\`, the loader injected \`agent_timeout = 600.0\` into task metadata (\`rllm/tasks/loader.py\`, both \`_merge_task_toml_metadata\` and \`_load_task_from_dir\`). That phantom default silently caps the agent's wall-clock budget at 600s and makes an operator-set \`RLLM_HARNESS_RUN_TIMEOUT_S\` ineffective:

- On the CLI harness (opencode/claude-code/etc.), the effective budget is \`min(agent_timeout, run_timeout)\` — so \`min(600, 3600) = 600\` no matter what the operator sets.
- Symptom in the wild: opencode SWE rollouts (scaleswe, whose task specs don't declare \`timeout_sec\`) getting SIGKILLed at ~660s (600 budget + 60s harness grace), even with \`RLLM_HARNESS_RUN_TIMEOUT_S=3600\`.

## Fix

Only set \`agent_timeout\` when the task actually declares \`[agent].timeout_sec\`. An absent value now stays absent, and the downstream consumers already do the right thing with it:
- \`cli_harness._effective_timeout\`: \`if per_task is None: return run_timeout\`
- \`eval/_resolution.py\`: \`if _per_task is None: agent_t = run_cap or 3600\`

So the operator's \`RLLM_HARNESS_RUN_TIMEOUT_S\` (or its default) governs for tasks that don't declare their own budget, instead of a hidden 600s floor. Tasks that *do* declare \`timeout_sec\` are unchanged.

Other readers keep their own explicit fallbacks (\`bash.py\` → 600, \`oracle.py\` → 7200), so behavior there is unchanged.

## Test

\`tests/tasks/test_loader_agent_timeout.py\` — undeclared → key omitted (fails without the fix), declared → preserved, base value survives a silent \`task.toml\`.

### DIFF
--- a/rllm/tasks/loader.py
+++ b/rllm/tasks/loader.py
@@ -230,7 +230,14 @@ def _merge_task_toml_metadata(task_dir: Path, base: dict) -> dict:
     merged["agent_user"] = raw.get("agent", {}).get("user", merged.get("agent_user"))
     merged["verifier_user"] = raw.get("verifier", {}).get("user", merged.get("verifier_user"))
     merged["verifier_timeout"] = raw.get("verifier", {}).get("timeout_sec", merged.get("verifier_timeout", 600.0))
-    merged["agent_timeout"] = raw.get("agent", {}).get("timeout_sec", merged.get("agent_timeout", 600.0))
+    # Only carry a per-task agent_timeout when the task actually declares one.
+    # No phantom default: an absent agent_timeout means "no per-task budget", so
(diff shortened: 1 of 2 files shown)


jeffreysijuntan:
Superseded — retargeting this fix onto terminal-rl instead (see the new PR).

PR #780, opened 2026-07-23
fix(eval): join Dockerfile RUN-continuation lines with a space, not a newline

### LINKED ISSUE #655: Dockerfile RUN-replay corrupts Harbor task setup on modal/daytona (line-continuation mangling + double-apply on prebuilt images)
## Summary

On non-docker sandbox backends (`modal`, `daytona`), the cold-path "replay the Dockerfile `RUN` steps" mechanism (`_replay_dockerfile` → `_dockerfile_run_commands` in `rllm/eval/_resolution.py`) corrupts Harbor / terminal-bench task setups in two distinct ways. Combined with the fact that these tasks boot from a **pre-built** per-task image, the replay also **double-applies** the build steps. The result is a flood of `Command failed in sandbox …` setup warnings and an unknown number of spuriously-failed (reward 0) tasks during eval.

Discovered while running `rllm eval <local harbor dir> --agent terminus2 --sandbox-backend modal`.

## Affected path

- `rllm/eval/_resolution.py:199` `_replay_dockerfile` (cold path; called by `_create_sandbox_for_task:213`)
- `rllm/eval/_resolution.py:251` `_dockerfile_run_commands`
- `rllm/eval/_resolution.py:298` `_resolve_image` (returns the configured pre-built image for non-docker backends)
- Snapshot builds are also affected: `build_modal_snapshot` / daytona snapshot build call the same `_replay_dockerfile`, so a snapshot bakes the same broken replay.
- `--sandbox-backend docker` is **not** affected — it builds the Dockerfile natively (`_replay_dockerfile` early-returns for `backend == "docker"`).

## Defect 1 — line-continuation mangling produces invalid shell

`_dockerfile_run_commands` strips the trailing `\` from each continued line and rejoins with `\n` (`rllm/eval/_resolution.py:278`):

```python
cmd = "\n".join(parts).strip()
```

A `\`-continuation in a Dockerfile `RUN` is a *shell line continuation* — the lines form **one** logical command. Joining with `\n` instead of a space turns a valid multi-line `&&` chain into invalid shell. `_replay_dockerfile` then execs the raw multi-line string via `_safe_exec` → `bash -c` (`:209-210`); note it does **not** use the existing `_as_single_run_line:285` helper.

### Reproduction

```python
from rllm.tasks.loader import BenchmarkLoader
from rllm.eval._resolution import _dockerfile_run_commands

br = BenchmarkLoader.load("<harbor tasks dir>", sandbox_backend="modal", harness_name="oracle")
t = next(x for x in br.tasks if x.id == "ansible-debug-project")
for c in _dockerfile_run_commands(t):
    print(repr(c))
```

For a Dockerfile containing:

```dockerfile
RUN apt-get update && apt-get install -y \
    python3-pip \
    && rm -rf /var/lib/apt/lists/*
```

the extracted "replay command" is:

```
apt-get update && apt-get install -y
    python3-pip
    && rm -rf /var/lib/apt/lists/*
```

which `bash` rejects with exactly the error seen in eval logs:

```
bash: -c: line 2: syntax error near unexpected token `&&'
```

## Defect 2 — `COPY`/`ADD` are dropped, and replay double-applies on a pre-built image

`_dockerfile_run_commands` intentionally keeps only `RUN` (docstring: *"only RUN is replayable on a live sandbox"*), dropping `COPY`/`ADD`. So replayed `RUN` steps that depend on copied files fail, e.g.:

```
cc1: fatal error: /tmp/vault.c: No such file or directory
python3: can't open file '/setup/generate_data.py': [Errno 2] No such file or directory
```

Crucially, on `modal`/`daytona` `_resolve_image` returns the **configured pre-built** image (e.g. `docker.io/<org>/optimbench-tb:<task>`), which is the *fully built* Dockerfile (FROM base + all RUN + COPY). Replaying the `RUN` steps on top of it re-runs the build → double-apply. Smoking gun, from a task whose Dockerfile has `RUN git clone https://github.com/JGCRI/hector.git /app/hector`:

```
Command failed in sandbox rllm-hector-carbon-budget-…: git clone https://github.com/JGCRI/hector.git /app/hector
stderr: fatal: destination path '/app/hector' already exists and is not an empty directory.
```

`/app/hector` can only "already exist" if the booted image already contains it — i.e. the image is the built Dockerfile and the replay is redundant.

## Impact

- Hundreds of `Command failed in sandbox …` warnings per eval run.
- Cold-path setup also balloons wall-clock (re-running the full build per task; observed ~15 min/task `setup=` on modal).
- An unknown subset of tasks score a spurious reward 0 when a non-idempotent re-run corrupts already-correct state. (Many tasks still pass because the pre-built image already has the right state, so the bug is often "just" noise — but it is not always harmless.)

## Suggested fixes

1. **Defect 1:** in `_dockerfile_run_commands`, join `\`-continuations with a single space (drop the `\`), matching shell semantics; or route every replay command through `_as_single_run_line`. This alone removes the `syntax error near '&&'` class.
2. **Defect 2 / double-apply:** when `_resolve_image` returns a configured pre-built image (rather than building the Dockerfile), skip `_replay_dockerfile` entirely — the image already contains the `RUN`+`COPY` result. Replay should only run when the booted image is the Dockerfile's `FROM` base. Alternatively, gate replay on a flag and prefer snapshots/native docker build for tasks whose Dockerfile uses `COPY`/`ADD`.

## Environment

- repo `rllm-org/rllm`, branch `terminal-rl`, commit `9833530`
- backend: `modal` (also applies to `daytona`); `docker` backend unaffected
- task format: Harbor / terminal-bench v2 tasks with multi-line `RUN` and/or `COPY` in `environment/Dockerfile`

AmirF194:
## Summary

`_dockerfile_run_commands` rejoins a Dockerfile `RUN` step's backslash-continued lines with `"\n".join(parts)` before `_replay_dockerfile` execs the result via `bash -c` on the modal/daytona sandbox backends. A `\`-continuation is a shell line continuation (one logical command), so joining with a newline instead of a space turns a valid multi-line `&&` chain into invalid shell and fails with `bash: -c: line N: syntax error near unexpected token '&&'`.

## Type of change

- [ ] Feature
- [x] Fix
- [ ] Docs
- [ ] Refactor
- [ ] Example / Project
- [ ] Infra / CI

## What changed

- `rllm/eval/_resolution.py`: `_dockerfile_run_commands` now joins continuation parts with `" ".join(part.strip() for part in parts)` instead of `"\n".join(parts)`.
- Added `tests/eval/test_dockerfile_replay.py`, a regression test that fails on main (asserts no embedded newline and that the extracted command parses as valid shell via `bash -n`) and passes with the fix.

## Validation

- [x] Targeted tests: `pytest tests/eval/test_dockerfile_replay.py tests/eval/test_resolver.py tests/sandbox/test_snapshot.py`
- [ ] Not run (reason below)

Validation details:
- Reproduced in a clean Docker container against HEAD `af40b7fa`: extracted the RUN step from a synthetic Dockerfile with a `\`-continuation, ran it through `bash -c`, got the exact syntax error from the issue. Applied the fix, same repro now runs clean (`returncode: 0`).
- New test fails on the pre-fix code (`assert "\n" not in commands[0]`) and passes after the fix, both run in the same container with `rllm.eval._resolution.__file__` printed to confirm it's exercising `/repo/rllm/eval/_resolution.py`, not a cached install.
- `pytest tests/eval/ tests/sandbox/test_snapshot.py` (excluding files that fail to collect in this scratch container for an unrelated reason, see below): 2 new + 35 existing pass, 1 pre-existing skip. 3 pre-existing failures in `test_snapshot.py` (`test_modal_ref_absent_confirmed_notfound_only`, `test_daytona_ref_absent_reads_enum_or_string_state`, `test_hook_install_honors_baked_script[cold-installs]`) reproduce identically on unpatched HEAD in the same container: `rllm_model_gateway` imports `fastapi.FastAPI`, which isn't installed in this minimal container. Unrelated to this change.
- `pre-commit run --files rllm/eval/_resolution.py tests/eval/test_dockerfile_replay.py`: ruff and ruff-format both pass.
- Did not run the full suite (`tests/eval/test_eval_engine.py` and others fail to collect in this container for the same missing `fastapi` reason) or exercise a real modal/daytona sandbox; the fix and test are scoped to the pure string-transform function.

- None

## Docs / examples

- [x] Not needed
- [ ] Updated docs
- [ ] Updated examples
- [ ] Follow-up docs needed

## Related issues / PRs

- Fixes #655
- Related to #656 (same fix, merged to `terminal-rl` instead of `main`; see note below)

## Screenshots / logs

None.

## Note on prior art

#656 fixed this exact join bug (and a separate double-apply-on-a-prebuilt-image defect from the same issue, via a new `[environment].replay_dockerfile` toggle) but its base branch was `terminal-rl`, not `main`, so `main` never picked it up and #655 stayed open. This PR ports only the line-continuation fix to `main`; the toggle for the double-apply defect is a separate concern and out of scope here.

### DIFF
--- a/rllm/eval/_resolution.py
+++ b/rllm/eval/_resolution.py
@@ -275,7 +275,7 @@ def _dockerfile_run_commands(task: Task) -> list[str]:
                 if i >= len(lines):
                     break
                 parts.append(lines[i])
-            cmd = "\n".join(parts).strip()
+            cmd = " ".join(part.strip() for part in parts).strip()
             if cmd:
                 commands.append(cmd)
         i += 1
--- a/tests/eval/test_dockerfile_replay.py
+++ b/tests/eval/test_dockerfile_replay.py
@@ -1,0 +1,46 @@
+"""Tests for :func:`rllm.eval._resolution._dockerfile_run_commands`.
+
+Non-docker sandbox backends (modal, daytona) replay a task Dockerfile's
+``RUN`` steps by extracting each step as a shell command and exec'ing it
+via ``bash -c`` (see ``_replay_dockerfile``). A multi-line ``RUN`` with
+``\\``-continuations must collapse into one valid shell command.
+"""
+
+from __future__ import annotations
+
+import subprocess
+from pathlib import Path
+
+from rllm.eval._resolution import _dockerfile_run_commands
+from rllm.types import Task
+
+
+def _task_with_dockerfile(tmp_path: Path, dockerfile_body: str) -> Task:
+    env_dir = tmp_path / "environment"
+    env_dir.mkdir()
+    (env_dir / "Dockerfile").write_text(dockerfile_body)
+    return Task(id="t", instruction="", metadata={}, dataset_dir=tmp_path)
+
+
+def test_multiline_run_continuation_collapses_to_one_valid_shell_command(tmp_path):
+    task = _task_with_dockerfile(
+        tmp_path,
+        "FROM python:3.11-slim\nRUN apt-get update && apt-get install -y \\\n    python3-pip \\\n    && rm -rf /var/lib/apt/lists/*\n",
+    )
+
+    commands = _dockerfile_run_commands(task)
+
+    assert len(commands) == 1
+    assert "\n" not in commands[0]
+    # The extracted command must parse as a single valid shell command
+    # (bash -n only checks syntax, no execution).
+    result = subprocess.run(["bash", "-n", "-c", commands[0]], capture_output=True, text=True)
+    assert result.returncode == 0, result.stderr
+
+
+def test_single_line_run_step_is_unaffected(tmp_path):
+    task = _task_with_dockerfile(tmp_path, "FROM python:3.11-slim\nRUN echo hello\n")
+
+    commands = _dockerfile_run_commands(task)
+
+    assert commands == ["echo hello"]


jeffreysijuntan:
Thanks for your PR. If you are working on sandboxed related agentic RL training task, feel free to switch over to the `terminal-rl` branch where we are actively developing things there. We plan to merge that into main soon after testing so this will get merged in along with it.

AmirF194:
Makes sense, happy to have this land as part of the terminal-rl merge rather than standalone. I flagged the same #656 overlap in the PR description, so whichever path is less work on your end is fine here.

AmirF194:
For visibility, since this landing via terminal-rl: #791 has the same fix directly against main (join with a space instead of newline), in case that's a simpler path than waiting for the branch merge.

AmirF194:
No rush, just checking in since it's been about two weeks. Happy to keep this parked until the terminal-rl merge, or land it standalone now if that's easier on your end.

jeffreysijuntan:
We will merge your PR directly since the `terminal-rl` branch is scheduled to merge later in the week. Thanks for the great work!

AmirF194:
Appreciate it, thanks for merging directly instead of waiting on the branch.

PR #797, opened 2026-07-25
fix(gateway): tombstone deleted sessions to stop trace-store leak

jeffreysijuntan:
## Problem

The gateway's in-memory trace store (`--store memory`) leaks sessions during async RL training. Observed live on a Fireworks run: **~157 orphaned sessions/hr**, each exactly **1 trace**, at old weight versions, with large mid-episode prompts — accumulating unbounded (the store has no TTL/eviction).

## Root cause — a persist-after-delete race

Every rollout correctly deletes its own gateway session in a `finally` (so filtering, staleness, and errors are all cleaned up — those were *not* the leak). The leak is a lifecycle race:

1. Traces persist **fire-and-forget** (`sync_traces=False`): the proxy does `asyncio.create_task(store_trace(...))` after each completion returns.
2. When a rollout ends it calls `flush()` then `delete_session(uid)`. `flush()` only drains persist tasks that *already exist* — an LLM call still in flight at the provider hasn't reached the persist line yet, so it can't be flushed.
3. That straggler returns **after** the delete → `store_trace` runs → and `MemoryTraceStore` **resurrects the session** because `self._session_index[session_id]` is a `defaultdict(list)` with no tombstone. Nothing deletes that `uid` again → permanent orphan with exactly 1 trace.

Signature match: single trace (the lone straggler), large prompt (slow `_build_trace_data` widens the window under high concurrency), grouped by task-uuid (a hard prompt makes several group members straggle together), early weight version (early policy aborts right after big tool-call turns).

## Fix — a permanent tombstone

Session ids are unique uuids (`uid = f"{uuid4()}:{rollout_idx}"`) that are **never reused**, so a deleted session must be rejected **permanently**, not for a time window:

- `delete_session` records the `session_id` in a tombstone set.
- `store_trace` drops any write for a tombstoned session instead of resurrecting it — always correct, since a deleted `uid` is never legitimately written again.
- The tombstone set is FIFO-capped (`max_tombstones`, default **1,000,000**) purely as a memory backstop for long-lived multi-run gateways. The cap is far larger than the number of sessions deletable while any request is still in flight, so eviction never exposes a live straggler. `max_tombstones=0` disables the guard (legacy behaviour).

Why not require pre-creation instead? Sessions are **implicitly created on first request** (a documented gateway feature for captured/external sessions), and the persist path (`ReverseProxy`) doesn't hold the `SessionManager`'s liveness state — so remembering *deletions* (a tombstone) is the correct store-level mechanism.

Why not a TTL? A time window has a real failure mode — if a request outlives it (the heartbeat budget allows up to 3600s), a very-late straggler could still resurrect the session. A permanent block has no such edge.

Cost: one interned `session_id` per deleted session (~tens of MB worst-case for an entire multi-day run, hard-capped); no change to the read/serialization path.

## Tests

`tests/unit/test_store.py` — new `TestTombstone`: straggler-after-delete is dropped (not resurrected), the block is permanent across repeated late writes, unrelated sessions unaffected, FIFO cap evicts the oldest tombstone, and `max_tombstones=0` restores legacy behaviour. Full store suite: **37 passed**.

## Scope / follow-up

`SqliteTraceStore` shares the same resurrection pattern (`INSERT OR REPLACE`/`IGNORE` on write, plain `DELETE`) and would take the identical guard — left as a follow-up since the observed leak is on the in-use `memory` backend.

This is a separate fix from the array-packing perf change (#787, already merged into this branch); the leak is a distinct, slower, unbounded issue on top of the O(N²) working-set growth.

### DIFF
--- a/rllm-model-gateway/src/rllm_model_gateway/store/memory_store.py
+++ b/rllm-model-gateway/src/rllm_model_gateway/store/memory_store.py
@@ -53,33 +53,36 @@ def _unpack(data: dict[str, Any]) -> dict[str, Any]:
 class MemoryTraceStore:
     """Ephemeral in-memory store.  Useful for tests and short-lived processes."""
 
-    #: A session deleted within this many seconds is "tombstoned": a late
-    #: ``store_trace`` for it is dropped instead of resurrecting the session.
-    #: Traces persist fire-and-forget, so a completion still in flight when the
-    #: rollout ended can land *after* the trainer's ``delete_session``; without a
-    #: tombstone ``store_trace`` re-creates a ``defaultdict`` entry that nothing
-    #: ever deletes again, leaking the session for the process lifetime. Must
-    #: exceed the longest possible in-flight request. Set to 0 to disable.
-    _DEFAULT_TOMBSTONE_TTL_S = 1800.0
+    #: Cap on remembered tombstones. A deleted session is tombstoned
+    #: *permanently* — session ids are unique uuids that are never legitimately
+    #: reused, so any later write is a straggler that must never resurrect the
+    #: session. Traces persist fire-and-forget, so a completion still in flight
+    #: when the rollout ended can land *after* the trainer's ``delete_session``;
+    #: without a tombstone ``store_trace`` re-creates a ``defaultdict`` entry that
+    #: nothing ever deletes again, leaking the session for the process lifetime.
+    #: The cap is only a memory backstop for long-lived multi-run gateways: it is
+    #: far larger than the number of sessions deletable while any request is still
+    #: in flight, so FIFO eviction never exposes a live straggler. 0 disables.
+    _DEFAULT_MAX_TOMBSTONES = 1_000_000
 
-    def __init__(self, tombstone_ttl_s: float = _DEFAULT_TOMBSTONE_TTL_S) -> None:
+    def __init__(self, max_tombstones: int = _DEFAULT_MAX_TOMBSTONES) -> None:
         # trace_id -> data dict
         self._traces: dict[str, dict[str, Any]] = {}
         # trace_id -> created_at
         self._timestamps: dict[str, float] = {}
         # session_id -> list[trace_id]  (insertion order)
         self._session_index: dict[str, list[str]] = defaultdict(list)
-        # session_id -> deletion time; blocks post-delete resurrection
-        self._tombstones: dict[str, float] = {}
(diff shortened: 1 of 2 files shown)


AmirF194 (commented):
This repo's own retry path reuses the exact session id the tombstone is meant to permanently block, and I think that silently kills trace visibility on every retried rollout.

`process_task_with_retry` (rllm/engine/agentflow_engine.py:596-603) does this on retry:

```
for retry_attempt in range(1, self.retry_limit + 1):
    uid = f"{task_id}:{rollout_idx}"
    if retry_attempt > 1:
        await self.gateway.adelete_session(uid)
    ...
    episode = await self._run_single(task_obj, uid, ...)
```

`uid` is identical across attempts, and the comment above it says the delete is there so "the new attempt's enrich doesn't see a mix of trace records" (i.e. it expects the store to accept fresh writes for the same uid right after). With the new permanent tombstone in `memory_store.py:83` (`if session_id in self._tombstones: return`), that expectation breaks: once `delete_session(uid)` runs, every later `store_trace` for that same `uid` is dropped for good, including the retry attempt's own traces. `_run_single` (agentflow_engine.py:672, `traces = await self.gateway.aget_traces(uid)`) then feeds `_finish_episode` an empty trace list, so a retried rollout gets enriched with nothing.

I confirmed the divergence directly against `MemoryTraceStore` (no other deps needed), simulating exactly that retry sequence: store a trace for `uid`, `delete_session(uid)`, store a second trace for the same `uid`, then read it back.

- On main (before this PR, commit `43ad465`): the second trace is returned normally.
- On this PR's head (`ebe7f8a`): `get_session_traces(uid)` comes back `[]`.

That second case is also just what `TestTombstone.test_block_is_permanent` in this PR asserts as the intended behavior, so it isn't a corner the tests missed, it's that the "never reused" premise doesn't hold for this repo's own retry loop.

Non-blocking, since I don't know this codebase's retry volume in practice, but worth a look before merge given it'd be a silent training-signal loss rather than a crash. A scoped fix might be to have `delete_session` clear any existing tombstone before deleting when it's about to be reused for a retry (i.e. have the retry loop call something that un-tombstones `uid`), or key the tombstone off a per-attempt id instead of the bare `task_id:rollout_idx`.

jeffreysijuntan:
thanks for raising the point. Yeah I think there are some concerns around this approach so this PR wouldn't get merged.

PR #809, opened 2026-08-07
feat(sft,tinker): harden multi-turn SFT training

listar2000:
## Summary

This makes structured, multi-turn SFT a lossless and fail-closed path across rLLM's data boundary and hosted trainers. It normalizes reasoning/tool trajectories, aligns Qwen training and serving renders, preflights the exact hosted batch plan before spend, and hardens checkpoint, resume, optimizer, validation, and request-order semantics. Unsupported or ambiguous combinations fail locally with actionable errors.

## Type of change

- [x] Feature
- [x] Fix
- [x] Docs
- [x] Refactor
- [ ] Example / Project
- [x] Infra / CI

## What changed

- Canonicalize reasoning and tool calls without mutating source whitespace; preserve explicit masks under both think-tag import modes and reject ambiguous duplicate/inline thinking representations.
- Match Qwen's last-real-user reasoning boundary for string and canonical wrapped-tool-response content.
- Render every identity-order and planned shuffled/grouped train batch plus validation data before creating a hosted client.
- Isolate default Tinker output paths; require a versioned local/provider resume identity covering model, renderer/tokenizer, dataset, optimizer, and horizon; validate cursors before resumed-client creation.
- Restore one-step hosted submission overlap while draining before validation/checkpoint boundaries.
- Reject hosted-only override keys on verl with native alternatives; retain the documented structured-row capability edge.
- Pin a reproducible Fireworks SDK/cookbook pair and fail closed when a Fireworks checkpoint lacks its persisted raw-row cursor.

## Validation

- [x] Targeted tests
- [ ] Not run (reason below)

Validation details:
- Integrated the current `terminal-rl` tip (`e4c28795`) without conflicts.
- 245 focused trainer, schema, bridge, and dataset CLI tests passed on the pinned provider stack; the full trainer suite passed (162 tests).
- Changed-file pre-commit hooks, `uv lock --check`, `git diff --check`, and Mintlify validation passed.
- Independent data/verl and hosted-backend reviews approved the final schema, preflight, resume, cursor, pipelining, and provider API boundaries.
- Tinker E2E: Qwen3.5-4B LoRA completed 3 updates, validation at steps 0/2, a step-2 periodic checkpoint, final step-3 checkpoint, same-directory no-op resume, and local mismatch rejection.
- Fireworks E2E: Qwen3.5-9B LoRA completed 3 updates, validation at steps 0/2, resumable/final checkpoints, a READY promoted model, and provider-confirmed trainer deletion.
- One-step overlap was behaviorally tested and provider-compatible, but no throughput improvement is claimed because throughput was not measured.

- `rllm dataset import --trim-whitespace` is removed. Core import now preserves source whitespace; model-aware cleanup belongs at the renderer/dataset-recipe boundary.
- Assistant text beginning with a structural `<think>...</think>` block is rejected by the plain `messages` bridge; use `--format think-tags` or canonical thinking parts.
- Over-length hosted trajectories error by default. Set `data.rllm.overlength_policy=truncate` only when lossy compatibility is intentional.
- Final partial batches are trained, so epoch-derived step counts can increase by one.
- Reusing a Tinker `--output` path is explicit resume intent; legacy or identity-mismatched checkpoints require a new output directory.
- Fireworks auto-resume requires a positive persisted `dataloader.json` cursor; legacy/missing cursor state is rejected rather than replayed.
- `hf_template` remains unsupported on Tinker/Fireworks until native sampler-render parity exists; use `cumulative`/`stepwise`, or verl.

## Docs / examples

- [ ] Not needed
- [x] Updated docs
- [ ] Updated examples
- [ ] Follow-up docs needed

Updated `docs/guides/sft-distillation.mdx` with a concise import → inspect → short hosted run → scale workflow, backend support, preflight/resume behavior, config examples, and troubleshooting.

## Related issues / PRs

- Round 1 audit follow-up for this PR.

## Screenshots / logs

- Not applicable.

### DIFF
--- a/cookbooks/swe-rl/nemotron_fireworks_sft.yaml
+++ b/cookbooks/swe-rl/nemotron_fireworks_sft.yaml
@@ -1,0 +1,57 @@
+# Fireworks port of the measured Nemotron 3 Nano SWE-agent SFT run.
+#
+# This file is merged over rllm/trainer/sft/config/fireworks.yaml.  Keep the
+# Fireworks model, Hugging Face tokenizer, and training shape in lockstep.
+
+model:
+  name: accounts/fireworks/models/nemotron-nano-3-30b-a3b
+  tokenizer_model: nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16
+  lora_rank: 32
+  # fireworks-ai 1.2.1 defaults all three targets to true.  The pinned rLLM
+  # backend does not forward these fields, so the preflight verifies the SDK
+  # defaults before any paid trainer is provisioned.
+  train_unembed: true
+  train_attn: true
+  train_mlp: true
+
+fireworks_config:
+  policy_trainer_shape_id: accounts/fireworks/trainingShapes/nemotron-nano-3-30b-a3b-262k-b200-lora
+  policy_trainer_replica_count: 1
+
+data:
+  train_batch_size: 9
+  micro_batch_size_per_gpu: 9
+  train_max_samples: -1
+  val_max_samples: -1
+  max_length: 65536
+  renderer_name: nemotron3
+  rllm:
+    tokenize_and_mask_method: cumulative
+    group_by_length: false
+    length_group_factor: 50
+    overlength_policy: error
+    # Match the measured run: one cross-entropy mean over all supervised
+    # assistant tokens in the nine-trajectory optimizer batch.
+    loss_reduction: token_mean
+    strip_thinking_from_history: false
+
+optim:
+  lr: 1.0e-4
+  betas: [0.9, 0.999]
+  eps: 1.0e-8
+  lr_scheduler: cosine
+  warmup_steps_ratio: 0.0
+  warmup_steps: 50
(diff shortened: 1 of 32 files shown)


signalrush:
This draft is now superseded for review by a native stacked-PR chain:

1. #841 Parquet container preservation
2. #842 reasoning/tool-call normalization
3. #843 think-mask preservation
4. #844 structural-thinking guardrail
5. #845 gateway tool-call serialization
6. #846 renderer wire-history normalization
7. #847 SFT/serving renderer parity
8. #848 forward-only validation
9. #849 completed-step validation boundaries
10. #850 batching and loss semantics
11. #851 exact training horizon
12. #852 optimizer controls
13. #853 provider preflight

Each PR targets the preceding branch and therefore shows only its own focused diff. Merge bottom-up from #841. I am leaving #809 open as historical context while the stack is reviewed. Resume hardening, performance, provider-version maintenance, and the experiment recipe will remain separate follow-up stacks.

PR #812, opened 2026-08-07
feat(eval): honour harbor's separate verifier environment (opt-in)

signalrush:
**New feature**, off by default behind `RLLM_SEPARATE_VERIFIER_ENV=1`.

⚠️ **Contains #811** (grading correctness) as its first 4 commits — it builds on the `[verifier]` contract parsing added there. Review #811 first; this diff is only the last 3 commits (+341/−20). Happy to rebase once #811 lands.

Implements harbor's separate-verifier contract, which rLLM tasks already declare but nothing acted on: `[verifier].environment_mode`, `[verifier.environment]`, `[[verifier.collect]]`, and top-level `artifacts` appear in every DeepSWE `task.toml` and in no code path.

## What it does

Sequence mirrors harbor's `trial/single_step.py`: run `[[verifier.collect]]` in the agent's box → pull the artifacts it snapshotted → create a fresh container → re-materialise them at their original paths → grade there → tear it down.

The verifier box is created from the task's image and `tests/` is uploaded to `/tests` as the evaluator already did, which reproduces what harbor builds from `tests/` as a build context (DeepSWE's `tests/Dockerfile` is "the pinned task image with the hidden tests baked in"). Resources come from `[verifier.environment]`.

The HEAD workaround from #811 goes **inert** here — a fresh box is already at the image's commit, which is the assumption that reset existed to fake.

## `download_file` on the Sandbox protocol

`Sandbox` had `upload_file`/`upload_dir` and no counterpart, so nothing could leave a sandbox. Harbor treats this as a first-class environment primitive (`download_file`/`download_dir` abstract on `BaseEnvironment`), and its `tar_transfer` module spells out why shelling out is wrong: per-file writes "drop executable bits, symlinks, and empty directories", and a partial transfer "can drop files silently".

| backend | mechanism |
|---|---|
| local | direct read |
| daytona | `fs.download_file` |
| docker | `container.get_archive` (tar) |
| modal | `filesystem.read_bytes`, base64-over-exec fallback |

Not `Sandbox.open()` on Modal — deprecated 2026-03-09 in favour of `Sandbox.filesystem`.

## Verified on Modal

| | result |
|---|---|
| separate mode (DeepSWE oracle, 4 tasks) | **4/4 = 1.0**, f2p 1.0, p2p 1.0, no deprecation warnings |
| shared mode (default, same 4 tasks) | **4/4** — unchanged |

Two bugs only end-to-end runs caught, both now covered by regression tests: the verifier container re-resolving its backend to `docker` while the agent ran on Modal, and the Modal `Sandbox.open()` deprecation.

## Why it's off by default

This is a behaviour change, not a bug fix. Honouring the contract is **stricter** than today's path: it grades `git diff <base> HEAD` (committed work only) where in-place grading sees the whole working tree, so it scores 0 for an agent that never commits. The oracle sweep in #811 shows in-place grading is not currently producing wrong verdicts (112/113), so this buys contract fidelity, tamper resistance and immunity to environment pollution — not a score correction. Enabling it by default should be a deliberate, separate decision.

## Two judgment calls worth a second opinion

1. **`[verifier.environment]` layers *over* `[environment]`** rather than replacing it — a deliberate deviation from harbor, which returns the verifier section wholesale. DeepSWE declares only cpus/memory/storage there and no image, so replacing would leave nothing to boot.
2. **Artifacts stage through a temp file** on the way in, since the protocol has no "write these bytes" primitive.

707 pass. One pre-existing failure, unrelated and untouched (`test_harbor_parity` lifetime, drift from #736).

### DIFF
--- a/rllm/engine/agentflow_engine.py
+++ b/rllm/engine/agentflow_engine.py
@@ -841,7 +841,13 @@ class AgentFlowEngine:
         for signal in eval_output.signals:
             enriched.metrics[signal.name] = signal.value
 
-        if _no_usable_model_output(enriched) and not enriched.is_correct and enriched.termination_reason not in INFRA_ERROR_REASONS:
+        # Harnesses that drive no model at all (oracle: runs the task's reference
+        # solution) trip the canary below by construction — every rollout has zero
+        # completions because none were ever requested. Left unguarded it reports
+        # the one thing the oracle exists to detect, a task whose *reference
+        # solution* fails its own verifier, as "upstream/proxy/gateway failure".
+        _llm_driven = getattr(self.agent_flow, "makes_llm_calls", True)
+        if _llm_driven and _no_usable_model_output(enriched) and not enriched.is_correct and enriched.termination_reason not in INFRA_ERROR_REASONS:
             # The agent produced nothing usable — no LLM calls at all, or every
             # call came back empty: a downed/erroring upstream (proxy died, auth or
             # tunnel failure), NOT a clean rollout. The reward (0) is meaningless;
--- a/rllm/engine/agentflow_engine.py
+++ b/rllm/engine/agentflow_engine.py
@@ -841,11 +841,9 @@ class AgentFlowEngine:
         for signal in eval_output.signals:
             enriched.metrics[signal.name] = signal.value
 
-        # Harnesses that drive no model at all (oracle: runs the task's reference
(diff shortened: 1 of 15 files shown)


signalrush:
Superseded — folded into a single PR. The HEAD restore alone was a workaround for grading in the wrong container and doesn't match harbor; the point is the contract-faithful implementation, which is now the path that runs.

PR #813, opened 2026-08-07
eval: honour harbor's verifier environment contract (reproduces DeepSWE 44%)

signalrush:
Supersedes #811 + #812, which shouldn't have been split: the HEAD restore alone is a *workaround* for grading in the wrong container, not a fix. Harbor does no git manipulation anywhere in its trial path because it never needs to. This is the contract-faithful implementation, with the workaround retained only as the fallback for in-place grading.

Builds on #810 (merged).

> **Status:** the 44% reproduction run against this exact commit is in flight; I'll post the result here. Everything below is measured.

## What rLLM was ignoring

Every DeepSWE `task.toml` declares a contract that appeared in no code path:

```toml
artifacts = ["/logs/artifacts/model.patch"]
[verifier]
environment_mode = "separate"
[[verifier.collect]]
command = "cd /app && ... git diff --binary <base_commit> HEAD > /logs/artifacts/model.patch"
```

rLLM ran the verifier in the agent's own container instead. The verifier restores official test files with `git checkout HEAD -- <path>` — a tamper guard assuming HEAD is the image's commit — while every DeepSWE instruction ends with *"work on this in a new branch from main and commit everything when you are done."* So HEAD moves, the restore resurrects the agent's own test files, the test patch no longer applies, the grader dies, and `test.sh`'s trap writes `-1` that was read as a legitimate reward.

## What this does

Resolves the mode exactly as harbor does (`harbor.models.task.verifier_mode._resolve_mode`): explicit `environment_mode` wins → a declared `[verifier.environment]` implies separate → everything else is shared. **The task's declaration is the only gate**, so this scopes itself to harbor SWE benchmarks with no global switch — verified across every materialized task locally:

```
terminal-bench (_hf_tasks)   shared    17792
deepswe                      separate    113
```

For separate-mode tasks the sequence mirrors harbor's `trial/single_step.py`: run `[[verifier.collect]]` in the agent's box → pull the artifacts → create a fresh container → re-materialise them at their original paths → grade there → tear it down. HEAD is the image's commit by construction, so the restore is inert.

For shared-mode tasks nothing changes, and the HEAD restore is scoped to separate-mode tasks graded in place — which also fixes a latent corruption: terminal-bench's `git-object-builder` ships a repo to repair and is scored on `refs/heads/main` and `git ls-tree HEAD`, so a blanket reset destroyed the deliverable. It is no longer probed at all.

## `download_file` on the Sandbox protocol

`Sandbox` had upload primitives and no counterpart, so nothing could leave a sandbox. Harbor treats this as first-class (`download_file`/`download_dir` abstract on `BaseEnvironment`), and its `tar_transfer` module explains why shelling out is wrong: per-file writes "drop executable bits, symlinks, and empty directories", and a partial transfer "can drop files silently".

Both directions are now native on every backend:

| backend | download | upload |
|---|---|---|
| local | read | `shutil.copy2` |
| daytona | `fs.download_file` | `fs.upload_file` |
| docker | `get_archive` (tar) | `put_archive` (tar) |
| modal | `filesystem.read_bytes` | `filesystem.copy_from_local` |

Modal's `upload_file` had been on base64-over-exec behind a stale "file API is in alpha" comment; `upload_dir` now ships its tar via `filesystem.write_bytes` and extracts in place. base64 remains the fallback on both paths. Not `Sandbox.open()` — deprecated 2026-03-09.

## Also here

- **Crash sentinel tagged** — a negative reward is "the grader died", not a score → `VerifierCrashError` → `GRADING_ERROR`, so training filters it instead of learning from a spurious zero.
- **Verifier stdout kept on the episode.** Harbor verifiers deliberately cat every suite log to stdout so "the reason a test failed is never lost", and the sandbox dies at teardown — a failed task recorded only `f2p 42/43` with no way to see which test. This identified an unwinnable task's cause on the first try (10 pass-to-pass tests binding IPv6 loopback, unavailable on Modal).
- **Empty-completion canary gated on `makes_llm_calls`** — the oracle drives no model, so every episode tripped it, reporting the one thing the oracle exists to find (a reference solution failing its own verifier) as "upstream/proxy/gateway failure".

## Evidence

An **oracle sweep** — every task's reference `solution/solve.sh` through the same grader, no LLM — scores **112/113 = 99.1%**. Cheap (~15 min, zero LLM tokens) and the fastest way to separate "harness broken" from "model missed". The one failure is unwinnable on Modal, not an rLLM defect (IPv6 loopback; the dataset declares `default_sandbox = "docker"`, which supplies `::1`).

Separate-mode grading verified end-to-end on Modal: **4/4 = 1.0**, f2p 1.0, p2p 1.0. Shared mode unchanged on the same tasks: 4/4.

Two bugs only end-to-end runs caught, both now regression-tested: the verifier container re-resolving its backend to `docker` while the agent ran on Modal, and the Modal `Sandbox.open()` deprecation.

**On reading a single eval run:** two full 113-task runs pooled to 98/226 = 43.4% (43.75% excluding the unwinnable task), but individually they were 40.7% and 46.0% — McNemar p=0.377, with **32 of 113 tasks (28%) flipping between runs** at temperature 1.0. A single n=113 run cannot resolve 40% from 46%.

## Judgment call worth a second opinion

**`[verifier.environment]` layers *over* `[environment]`** rather than replacing it — a deliberate deviation from harbor, which returns the verifier section wholesale. DeepSWE declares only cpus/memory/storage there and no image, so replacing would leave nothing to boot.

## Testing

708 pass. One pre-existing failure, unrelated and untouched: `test_harbor_parity.py::test_lifetime_floor_sized_from_task_budget_when_no_override` expects `agent+verifier+install+slack` (3000) while `_sandbox_resource_kwargs` computes `agent+verifier+300` (2100) — drift from #736. Its intent (counting install time toward sandbox lifetime) is arguably right, which makes it a latent reap risk for slow-install benchmarks rather than a stale assertion.

### DIFF
--- a/rllm/engine/agentflow_engine.py
+++ b/rllm/engine/agentflow_engine.py
@@ -841,7 +841,13 @@ class AgentFlowEngine:
         for signal in eval_output.signals:
             enriched.metrics[signal.name] = signal.value
 
-        if _no_usable_model_output(enriched) and not enriched.is_correct and enriched.termination_reason not in INFRA_ERROR_REASONS:
+        # Harnesses that drive no model at all (oracle: runs the task's reference
+        # solution) trip the canary below by construction — every rollout has zero
+        # completions because none were ever requested. Left unguarded it reports
+        # the one thing the oracle exists to detect, a task whose *reference
+        # solution* fails its own verifier, as "upstream/proxy/gateway failure".
+        _llm_driven = getattr(self.agent_flow, "makes_llm_calls", True)
+        if _llm_driven and _no_usable_model_output(enriched) and not enriched.is_correct and enriched.termination_reason not in INFRA_ERROR_REASONS:
             # The agent produced nothing usable — no LLM calls at all, or every
             # call came back empty: a downed/erroring upstream (proxy died, auth or
             # tunnel failure), NOT a clean rollout. The reward (0) is meaningless;
--- a/rllm/engine/agentflow_engine.py
+++ b/rllm/engine/agentflow_engine.py
@@ -841,11 +841,9 @@ class AgentFlowEngine:
         for signal in eval_output.signals:
             enriched.metrics[signal.name] = signal.value
 
-        # Harnesses that drive no model at all (oracle: runs the task's reference
-        # solution) trip the canary below by construction — every rollout has zero
-        # completions because none were ever requested. Left unguarded it reports
-        # the one thing the oracle exists to detect, a task whose *reference
-        # solution* fails its own verifier, as "upstream/proxy/gateway failure".
+        # A harness that drives no model (oracle) has zero completions by
+        # construction, so the canary below would report its every failure — i.e.
+        # a task whose reference solution fails its own verifier — as an outage.
         _llm_driven = getattr(self.agent_flow, "makes_llm_calls", True)
         if _llm_driven and _no_usable_model_output(enriched) and not enriched.is_correct and enriched.termination_reason not in INFRA_ERROR_REASONS:
             # The agent produced nothing usable — no LLM calls at all, or every
--- a/rllm/eval/_resolution.py
+++ b/rllm/eval/_resolution.py
@@ -20,6 +20,7 @@ import base64
 import importlib
 import inspect
 import logging
+import os
 import re
 import shlex
 import uuid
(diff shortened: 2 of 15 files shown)


signalrush:
## Reproduction result

Ran the full benchmark against this branch (`afd34199`), graded under the harbor contract this PR implements:

```
rllm eval deepswe --agent mini-swe-agent --sandbox-backend modal \
  --model accounts/fireworks/models/glm-5p2 --concurrency 24
```

**50/113 = 44.2%** — DeepSWE's published figure for glm-5p2 is 44%. (SE 4.7%, so 44% sits 0.05 SE from the point estimate.)

| | |
|---|---|
| terminations | 109 `env_done`, 2 `timeout`, 1 `error`, 1 `verifier_timeout` |
| signal averages | f2p 0.876, **p2p 0.993** |
| gateway throttles | 23 across ~13k model calls, 0 leaked under a committed 200 |
| verifier-provisioning failures | **0** |

### The contract path works with a real agent

The open question going in was whether mini-swe-agent commits reliably enough for separate grading — the contract carries work as `git diff <base_commit> HEAD`, so uncommitted edits don't count and a non-committing harness would score 0 everywhere. It does: **zero tasks showed the empty-patch signature** (`f2p == 0` with the grader reporting "no model.patch submitted"). Failures are ordinary near-misses instead — the first one graded was f2p 73/78 with real assertion errors on the remaining 5.

p2p averaging 0.993 confirms the fresh verifier containers are healthy.

### The 4 non-`env_done` outcomes, checked individually

- **1 `error`** — a model call hung until the gateway's 3600s heartbeat budget expired, killing the agent (exit 143). Upstream latency, unrelated to this PR; the verifier container is only created after the agent finishes.
- **1 `verifier_timeout`** — `pwntools-tube-multiplexing` ran 1807s against its declared 1800s budget. The oracle grades that task in 15–23s, so the verifier isn't slow; the agent's tube-multiplexing implementation hangs the suite. Correctly caught.
- **2 `timeout`** — agents out of wall-clock.

### Reading this number

One run can't certify 44% on this benchmark. Two earlier full runs (in-place grading) gave 40.7% and 46.0%, with **32 of 113 tasks (28%) flipping between them** and McNemar p=0.377. At n=113 the 95% CI here is 35.1–53.4%. So the honest claim is that this commit produced 44.2% and that DeepSWE's 44% is indistinguishable from it — not that any single run pins the number.

Also worth noting for anyone reproducing: `--concurrency` defaults to 64, which produced sustained 30–70 Fireworks 429s/min and starved 3 tasks into wall-clock timeouts (one a genuine solve — it passed at concurrency 4 with 118 steps vs 24 steps at 64). #810 keeps those throttles from corrupting completions, but the backoff still eats the agent's budget, so 24 is the safer setting.

PR #820, opened 2026-08-10
fix(gateway): record Anthropic /v1/messages traces; unpin litellm past the thinking_blocks bug

Chanbinski2:
Two provider/wire-format fixes, independent of any one benchmark. Split out from the TB3 work so they can be reviewed on their own merits.

## 1. The gateway could not read Anthropic responses

`build_trace_record` reads `choices[0].message`, which only exists in the OpenAI shape. The `claude-code` harness drives the CLI over `ANTHROPIC_BASE_URL`, so its traffic is Anthropic-shaped — the gateway's `/v1/{path:path}` catch-all forwarded it fine, but every trace came back empty.

Running both versions against a real Anthropic response body:

```
before:  response_message={}  finish_reason=None  token_counts={}
after:   response_message={role, content, tool_calls}
         finish_reason=tool_calls  token_counts={prompt: 17664, completion: 82}
```

Eval still scored correctly, because reward comes from the verifier — which is why this went unnoticed. But trajectories were empty, so there was **no training signal at all** from an Anthropic-speaking harness.

Handles both directions:

- **non-streaming** — `type == "message"` with `content[]` blocks folded into content/`tool_calls`, `stop_reason` → `finish_reason`, `input_tokens`/`output_tokens` → prompt/completion
- **streaming** — `message_start` / `content_block_start` / `content_block_delta` / `message_delta`, with `input_json_delta` accumulated per block index so parallel tool calls reassemble correctly

Usage is merged rather than overwritten, since Anthropic splits it across `message_start` and `message_delta`.

Generic to the wire format, so it also covers `aider`, `opencode`, and `mini_swe_agent` whenever they are pointed at `ANTHROPIC_BASE_URL`.

## 2. litellm 1.83.0 breaks multi-turn tool use on strict providers

litellm converts the provider's `reasoning_content` into its own `thinking_blocks` field, then forwards that field upstream on the next request. Fireworks rejects unknown message fields:

```
400 invalid_request_error:
  Extra inputs are not permitted, field: 'messages[2].thinking_blocks'
```

Claude Code's second request — the one echoing the assistant's reasoning back alongside the tool result — always carries it, so **every rollout died at turn 2, on every task**. Measured on a real eval:

| litellm | turns | tool calls | duration | is_error |
|---|---|---|---|---|
| 1.83.0 | 2 | 0 | 64s | true |
| 1.88.0 | 57 | 60 | 606s | false |

Confirmed three ways: an SDK-level control test on the identical message payload (1.83.0 400s, 1.88.0 strips the field), and end-to-end runs on both a Compose and a non-Compose task.

**Avoid 1.96.0** — its proxy imports `get_flat_dependant` from fastapi, removed in 0.140.13+, while its own declared bound is `fastapi<1.0,>=0.136.3`. 1.88.0, 1.91.0, and 1.93.2 were all verified working.

The `[tool.uv]` override is what binds here — bumping the `>=1.83.0` floor in dependencies alone has no effect (a plain `uv pip install` of a newer version is silently reverted).

## Testing

`rllm-model-gateway/tests/unit/test_data_process.py` — 24 passed. New cases cover Anthropic non-streaming, streaming, tool-call reassembly, and usage merging.

## Notes for reviewers

The litellm pin arrived in `8ce1946f` ("Add override-dependencies for sdk") with no stated reason for that specific version, so I've read it as incidental rather than a deliberate ceiling — worth confirming.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -143,7 +143,12 @@ rllm-model-gateway = { path = "./rllm-model-gateway", editable = true }
 
 [tool.uv]
 override-dependencies = [
-    "litellm[proxy]==1.83.0",
+    # >=1.88.0: 1.83.0 forwards its own ``thinking_blocks`` field to the
+    # upstream provider on the Anthropic /v1/messages path. Providers that
+    # reject unknown message fields (Fireworks) 400 the *second* request of
+    # every CLI-harness rollout — the one echoing the assistant's reasoning
+    # back with the tool result — so agents die at turn 2 on every task.
+    "litellm[proxy]>=1.88.0",
     "numpy>=1.26.0",
     "rich>=14.1.0",
 ]
--- a/rllm-model-gateway/src/rllm_model_gateway/data_process.py
+++ b/rllm-model-gateway/src/rllm_model_gateway/data_process.py
@@ -1,10 +1,11 @@
-"""Token ID / logprob extraction from OpenAI-style responses.
+"""Token ID / logprob extraction and trace construction for model responses.
 
 Extracted from ``rllm/sdk/data_process.py``.  No dependency on rLLM's
 ``ModelOutput``, ``Step``, or ``Trajectory`` types — only operates on plain
 dicts and produces ``TraceRecord`` instances.
 """
 
+import json
 import logging
 import time
 import uuid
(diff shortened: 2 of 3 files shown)


Chanbinski2:
Superseded — combining into a single PR.

PR #821, opened 2026-08-10
feat(eval): native Terminal-Bench 3 runtime — Compose environments, Harbor 0.20, verifier contract

Chanbinski2:
Runs Terminal-Bench 3 natively in rLLM's own sandbox/eval stack, with fidelity to Harbor's task contract, rather than delegating to Harbor's runner.

Stacked on #820 (gateway Anthropic traces) — see *Dependency* below.

## Harbor 0.20 / TB3 packaging

TB3 ships as a Hub package (`terminal-bench/terminal-bench@3.0.0`, the `org/name@ref` form) rather than a legacy `name@version` dataset. `get_harbor_dataset_client()` dispatches between `PackageDatasetClient` and `RegistryClientFactory`, mirroring Harbor's own CLI, so both forms keep working. Bumps the `harbor` extra 0.3.0 → 0.20.0 and adds `docker>=7.0.0`.

## Dataset registration

Adds `terminal-bench-3` to the registry — 74 tasks, `default_agent: claude-code`.

## Docker Compose environments

Some TB3 tasks ship `environment/docker-compose.yaml` with multiple services, which no existing backend can run. Adds `DockerComposeSandbox` and `ModalComposeSandbox`; `_create_sandbox_for_task` routes to them when a task has a compose file and raises a clear error on backends that cannot.

Snapshotting is skipped for compose tasks — a VM plus nested images, networks, and volumes has no single-image snapshot representation.

The sandbox-backend edits here are **prerequisites, not standalone cleanups**, which is why they aren't in #820:

- `docker_compose` needs `DockerSandbox.from_container()` to attach the Sandbox file/exec API to a Compose-owned container
- `modal_compose` needs `shell`, `experimental_options`, and `block_network` on `ModalSandbox`

`DockerSandbox` also gains cpu/memory/gpu limits from the task's environment section.

## Verifier contract fidelity

Harbor builds the verifier image *from* `tests/` as build context, and that Dockerfile may install dependencies or transform files — uploading raw `tests/` into the agent image is not equivalent. `_create_verifier_sandbox` now builds `tests/Dockerfile` per backend (docker/modal/daytona), with `verifier_tests_baked` and `verifier_env_vars` plumbed through. Tasks predating the `tests/Dockerfile` contract keep the old path.

## Testing

58 passed across the new and touched suites (`test_docker_compose`, `test_modal_compose`, `test_snapshot`, `test_modal_backend`, `test_terminal_bench_3`, `test_terminal_bench_3_runtime`, `test_claude_code_tb3`).

One pre-existing failure, `test_snapshot.py::test_daytona_ref_absent_reads_enum_or_string_state`, fails identically on `terminal-rl` without this branch — the optional `daytona` dep isn't installed. Untouched by this PR.

End-to-end, both task shapes:

| task | compose | turns | outcome |
|---|---|---|---|
| `freight-dispatch-shift` | ✅ | 176 | artifact collected, verifier graded 132/232 diagnostic points |
| `html-js-filter` | ❌ | 60 | 1 of 2 verifier tests passed |

Each task's native `task.toml` budgets were honoured without `--agent-timeout` — 7200s/300s and 3600s/1800s respectively.

## Dependency

Scoring is correct without #820, but trajectories from the `claude-code` harness are empty until the gateway can parse Anthropic responses. Merge #820 first for non-empty training data.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -135,7 +135,8 @@ code-tools = [
 ]
 
 harbor = [
-    "harbor==0.3.0 ; python_version >= '3.12'",
+    "harbor==0.20.0 ; python_version >= '3.12'",
+    "docker>=7.0.0 ; python_version >= '3.12'",
 ]
 
 [tool.uv.sources]
--- a/rllm/cli/_pull.py
+++ b/rllm/cli/_pull.py
@@ -225,14 +225,14 @@ def resolve_harbor_catalog_entry(name: str) -> dict | None:
     a pre-registered entry in ``datasets.json``.
     """
     try:
-        from harbor.registry.client.factory import RegistryClientFactory
+        from rllm.integrations.harbor.dataset_loader import get_harbor_dataset_client
     except ImportError:
         return None
 
     import asyncio
 
     async def _probe():
-        client = RegistryClientFactory.create()
+        client = get_harbor_dataset_client(name)
         try:
             metadata = await client.get_dataset_metadata(name)
             return metadata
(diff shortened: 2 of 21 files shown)


Chanbinski2:
Superseded — combining into a single PR.

PR #822, opened 2026-08-10
feat(eval): native Terminal-Bench 3 runtime + Anthropic trace/litellm infra fixes

Chanbinski2:
Runs Terminal-Bench 3 natively in rLLM's own sandbox/eval stack, with fidelity to Harbor's task contract, rather than delegating to Harbor's runner.

The change splits cleanly into **two independent parts**, kept as two commits so they can be reviewed (or reverted) separately.

---

# Part 1 — Infra: provider/wire-format fixes

Independent of any benchmark. `fc45b580`

### 1a. The gateway could not read Anthropic responses

`build_trace_record` reads `choices[0].message`, which only exists in the OpenAI shape. The `claude-code` harness drives the CLI over `ANTHROPIC_BASE_URL`, so its traffic is Anthropic-shaped — the gateway's `/v1/{path:path}` catch-all forwarded it fine, but every trace came back empty.

Running both versions against a real Anthropic response body:

```
before:  response_message={}  finish_reason=None  token_counts={}
after:   response_message={role, content, tool_calls}
         finish_reason=tool_calls  token_counts={prompt: 17664, completion: 82}
```

Eval still scored correctly, because reward comes from the verifier — which is why this went unnoticed. But trajectories were empty, so there was **no training signal at all** from an Anthropic-speaking harness.

Handles both directions:

- **non-streaming** — `type == "message"` with `content[]` blocks folded into content/`tool_calls`, `stop_reason` → `finish_reason`, `input_tokens`/`output_tokens` → prompt/completion
- **streaming** — `message_start` / `content_block_start` / `content_block_delta` / `message_delta`, with `input_json_delta` accumulated per block index so parallel tool calls reassemble correctly

Usage is merged rather than overwritten, since Anthropic splits it across `message_start` and `message_delta`.

Generic to the wire format, so it also covers `aider`, `opencode`, and `mini_swe_agent` whenever they are pointed at `ANTHROPIC_BASE_URL`.

### 1b. litellm 1.83.0 breaks multi-turn tool use on strict providers

litellm converts the provider's `reasoning_content` into its own `thinking_blocks` field, then forwards that field upstream on the next request. Fireworks rejects unknown message fields:

```
400 invalid_request_error:
  Extra inputs are not permitted, field: 'messages[2].thinking_blocks'
```

Claude Code's second request — the one echoing the assistant's reasoning back alongside the tool result — always carries it, so **every rollout died at turn 2, on every task**. Measured on a real eval:

| litellm | turns | tool calls | duration | is_error |
|---|---|---|---|---|
| 1.83.0 | 2 | 0 | 64s | true |
| 1.88.0 | 57 | 60 | 606s | false |

Confirmed three ways: an SDK-level control test on the identical message payload (1.83.0 400s, 1.88.0 strips the field), and end-to-end runs on both a Compose and a non-Compose task.

**Avoid 1.96.0** — its proxy imports `get_flat_dependant` from fastapi, removed in 0.140.13+, while its own declared bound is `fastapi<1.0,>=0.136.3`. 1.88.0, 1.91.0, and 1.93.2 were all verified working.

The `[tool.uv]` override is what binds — bumping the `>=1.83.0` floor in dependencies alone has no effect (a plain `uv pip install` of a newer version is silently reverted).

---

# Part 2 — Terminal-Bench 3 native runtime

`28a1c0c8`

### Harbor 0.20 / TB3 packaging

TB3 ships as a Hub package (`terminal-bench/terminal-bench@3.0.0`, the `org/name@ref` form) rather than a legacy `name@version` dataset. `get_harbor_dataset_client()` dispatches between `PackageDatasetClient` and `RegistryClientFactory`, mirroring Harbor's own CLI, so both forms keep working. Bumps the `harbor` extra 0.3.0 → 0.20.0 and adds `docker>=7.0.0`.

### Dataset registration

Adds `terminal-bench-3` to the registry — 74 tasks, `default_agent: claude-code`.

### Docker Compose environments

Some TB3 tasks ship `environment/docker-compose.yaml` with multiple services, which no existing backend can run. Adds `DockerComposeSandbox` and `ModalComposeSandbox`; `_create_sandbox_for_task` routes to them when a task has a compose file and raises a clear error on backends that cannot.

Snapshotting is skipped for compose tasks — a VM plus nested images, networks, and volumes has no single-image snapshot representation.

The sandbox-backend edits are **prerequisites, not standalone cleanups**, which is why they sit in Part 2 rather than Part 1:

- `docker_compose` needs `DockerSandbox.from_container()` to attach the Sandbox file/exec API to a Compose-owned container
- `modal_compose` needs `shell`, `experimental_options`, and `block_network` on `ModalSandbox`

`DockerSandbox` also gains cpu/memory/gpu limits from the task's environment section.

### Verifier contract fidelity

Harbor builds the verifier image *from* `tests/` as build context, and that Dockerfile may install dependencies or transform files — uploading raw `tests/` into the agent image is not equivalent. `_create_verifier_sandbox` now builds `tests/Dockerfile` per backend (docker/modal/daytona), with `verifier_tests_baked` and `verifier_env_vars` plumbed through. Tasks predating the `tests/Dockerfile` contract keep the old path.

---

## Testing

Gateway unit tests: 24 passed, with new cases for Anthropic non-streaming, streaming, tool-call reassembly, and usage merging.

Sandbox/TB3 suites: 58 passed. One pre-existing failure, `test_snapshot.py::test_daytona_ref_absent_reads_enum_or_string_state`, fails identically on `terminal-rl` without this branch — the optional `daytona` dep isn't installed. Untouched here.

End-to-end, both task shapes:

| task | compose | turns | outcome |
|---|---|---|---|
| `freight-dispatch-shift` | ✅ | 176 | artifact collected, verifier graded 132/232 diagnostic points |
| `html-js-filter` | ❌ | 60 | 1 of 2 verifier tests passed |

Each task's native `task.toml` budgets were honoured without `--agent-timeout` — 7200s/300s and 3600s/1800s respectively.

## Notes for reviewers

The litellm pin arrived in `8ce1946f` ("Add override-dependencies for sdk") with no stated reason for that specific version, so I've read it as incidental rather than a deliberate ceiling — worth confirming.

Part 2 scores correctly without Part 1, but `claude-code` trajectories stay empty until the gateway can parse Anthropic responses.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -135,7 +135,8 @@ code-tools = [
 ]
 
 harbor = [
-    "harbor==0.3.0 ; python_version >= '3.12'",
+    "harbor==0.20.0 ; python_version >= '3.12'",
+    "docker>=7.0.0 ; python_version >= '3.12'",
 ]
 
 [tool.uv.sources]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -143,7 +143,12 @@ rllm-model-gateway = { path = "./rllm-model-gateway", editable = true }
 
 [tool.uv]
 override-dependencies = [
-    "litellm[proxy]==1.83.0",
+    # >=1.88.0: 1.83.0 forwards its own ``thinking_blocks`` field to the
+    # upstream provider on the Anthropic /v1/messages path. Providers that
+    # reject unknown message fields (Fireworks) 400 the *second* request of
+    # every CLI-harness rollout — the one echoing the assistant's reasoning
+    # back with the tool result — so agents die at turn 2 on every task.
+    "litellm[proxy]>=1.88.0",
     "numpy>=1.26.0",
     "rich>=14.1.0",
 ]
--- a/rllm-model-gateway/src/rllm_model_gateway/data_process.py
+++ b/rllm-model-gateway/src/rllm_model_gateway/data_process.py
@@ -1,10 +1,11 @@
-"""Token ID / logprob extraction from OpenAI-style responses.
+"""Token ID / logprob extraction and trace construction for model responses.
 
 Extracted from ``rllm/sdk/data_process.py``.  No dependency on rLLM's
 ``ModelOutput``, ``Step``, or ``Trajectory`` types — only operates on plain
 dicts and produces ``TraceRecord`` instances.
 """
 
+import json
 import logging
 import time
 import uuid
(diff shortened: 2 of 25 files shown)


jeffreysijuntan:
For this new type of sandbox, can we have some sort of unified abstraction instead of different implementation for different backends? For example, have a base class called `DockerComposeSandbox` that inherits the original `Sandbox` class but add some extended functions to support this feature, and then each backend implements its own DockerComposeSandbox.

PR #830, opened 2026-08-12
Build the real Dockerfile on Modal instead of replaying RUN steps

thwu1:
## Summary

Modal was left out of `_FROM_DOCKERFILE_BACKENDS` when the real-Dockerfile build landed for Daytona in 9277e519, whose message noted: *"modal from_dockerfile is a tracked follow-up (untested here; `_FROM_DOCKERFILE_BACKENDS`)."* This finishes that follow-up.

On Modal the cold path pulled the Dockerfile's `FROM` base and replayed only its `RUN` steps, silently dropping `COPY`/`ENV`/`WORKDIR`. `WORKDIR` is what makes this fatal for harbor SWE tasks, whose environments do:

```dockerfile
WORKDIR /app
RUN git clone <url> .
```

Without `WORKDIR`, the clone lands in the base image's root, `/app` never exists, and every `cd /app` in `solution/solve.sh` and `tests/test.sh` fails. The verifier writes no reward file and the trial dies as `grading_error`:

```
Command failed (exit 1) in sandbox rllm-verify-...:
chmod +x /tests/test.sh && cd /app && /tests/test.sh
bash: line 1: cd: /app: No such file or directory
```

Found on a 992-task harbor SWE set where 522 tasks clone into `.` rather than an absolute path — none of them can run on Modal today.

## Why this is small

Modal already had every piece; only the wiring was missing.

| piece | status before |
|---|---|
| `modal.Image.from_dockerfile(path, context_dir=…)` | already in the SDK |
| `ModalSandbox` accepting a `modal.Image` object | already supported |
| `_dockerfile_image()` modal branch | missing — raised `ValueError` |
| `"modal"` in `_FROM_DOCKERFILE_BACKENDS` | missing |

`context_dir=dockerfile.parent` mirrors Daytona, whose `Image.from_dockerfile` infers the context from the file's directory, so `COPY` resolves identically on both.

## Snapshot path

`build_modal_snapshot` now mirrors `build_daytona_snapshot`: a from_dockerfile image already carries every `RUN`, so replaying would double-apply them, and a snapshotted task must come out identical to a cold-built one. Only the install script layers on top, and the replay term drops out of the build timeout.

`env_key` in `rllm/sandbox/snapshot.py` already routes through `_builds_from_dockerfile(task, backend)` generically, so context fingerprinting picks Modal up with no change there.

## Scope

Tasks declaring an explicit `docker_image` resolve `replay_dockerfile` to false and are **unaffected** — they still boot their prebuilt image as-is. Verified against two real tasks:

| task | `docker_image` | `should_replay` | builds Dockerfile on modal |
|---|---|---|---|
| prebuilt-image deepswe | `public.ecr.aws/…:…-v1.1` | `False` | No — unchanged |
| Dockerfile-only harbor task | none (`gcc:13` from `FROM`) | `True` | Yes |

This only changes tasks whose environment *is* their Dockerfile — which previously could not work on Modal at all.

## Tests

Four cases added to `tests/eval/test_dockerfile_replay.py`: Modal and Daytona both take the from_dockerfile path; `docker`/`local` do not; `replay_dockerfile = false` opts out on Modal; unsupported backends still raise.

`tests/eval` + `tests/sandbox`: **605 passed, 1 failed**. The failure is `test_harbor_parity.py::test_lifetime_floor_sized_from_task_budget_when_no_override`, which fails identically on `terminal-rl` without this change — pre-existing and unrelated.

## Not verified

The Modal from_dockerfile path has **not been exercised against live Modal** in this change — same caveat the original commit gave for its own untested Modal path. Static and unit-level verification only. A real `rllm eval --sandbox-backend modal` on a Dockerfile-based harbor task is the remaining check before relying on it.

### DIFF
--- a/rllm/eval/_resolution.py
+++ b/rllm/eval/_resolution.py
@@ -334,10 +334,8 @@ def _task_dockerfile(task: Task) -> Path | None:
 
 # Remote backends that build images themselves and so can build the *real* Dockerfile
 # (COPY/ENV/WORKDIR/RUN) instead of pulling FROM + replaying RUN. ``docker`` is excluded
-# because it already builds via ``docker build``; ``local`` cannot build. ``modal`` is a
-# tracked follow-up (it accepts a ``modal.Image`` and already keepalive-overrides the
-# entrypoint, but the from_dockerfile path there is untested — see _dockerfile_image).
-_FROM_DOCKERFILE_BACKENDS = ("daytona",)
+# because it already builds via ``docker build``; ``local`` cannot build.
+_FROM_DOCKERFILE_BACKENDS = ("daytona", "modal")
 
 
 def _builds_from_dockerfile(task: Task, backend: str) -> Path | None:
@@ -360,6 +358,18 @@ def _dockerfile_image(backend: str, dockerfile: Path):
         from daytona import Image  # daytona.Image.from_dockerfile bundles the Dockerfile dir as context
 
         return Image.from_dockerfile(str(dockerfile))
+    if backend == "modal":
+        import modal
+
+        # Modal takes the Dockerfile path and its build context separately; daytona
+        # infers the context from the file's directory, so pass the parent explicitly
+        # to keep COPY resolution identical across the two backends.
+        #
+        # WORKDIR is the directive that makes this matter beyond COPY: harbor SWE tasks
+        # (deepswe/uniswe) do `WORKDIR /app` then `git clone <url> .`, so under the
+        # replay path — which drops WORKDIR — the clone lands in the base image's root,
+        # /app never exists, and every `cd /app` in solve.sh and tests/test.sh fails.
+        return modal.Image.from_dockerfile(dockerfile, context_dir=dockerfile.parent)
     raise ValueError(f"from_dockerfile build unsupported for backend {backend!r}")
 
 
--- a/rllm/sandbox/backends/modal_backend.py
+++ b/rllm/sandbox/backends/modal_backend.py
@@ -568,7 +568,9 @@ def build_modal_snapshot(task, key: str, prior_ref: str | None = None, *, force:
     snapshot keyed on the install must actually contain it.
     """
     from rllm.eval._resolution import (
+        _builds_from_dockerfile,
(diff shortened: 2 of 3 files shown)


signalrush:
## Verified against live Modal

Ran the oracle harness on a Dockerfile-based harbor SWE task (`uniswe-ml/c-jsmn`, `FROM gcc:13`, `WORKDIR /app` + `git clone <url> .`) on Modal with this branch:

```
[uniswe-ml/c-jsmn:0] Rewards: [oracle: 1.0] in 9s (setup=3s agentflow=2s evaluator=3s teardown=0s)
Accuracy 100.0% (1/1)
verifier: reward=1 f2p_passed=38/38
solve.sh: [solve] reference solution applied via: git apply --whitespace=nowarn
```

**Before this change**, the same task on Modal:

```
Command failed (exit 1) in sandbox rllm-verify-uniswe-ml-c-jsmn-...:
chmod +x /tests/test.sh && cd /app && /tests/test.sh
bash: line 1: cd: /app: No such file or directory
→ termination_reason: grading_error, "no reward file found"
```

`/app` now exists because `WORKDIR` survives the real build. Modal's image cache also kicks in on re-run: setup went 25s (cold build) → 3s (cached).

### One caveat on the run, not on the change

Reaching `reward=1` also required `RLLM_SEPARATE_VERIFIER_ENV=0`. That is **not** a gap in this PR — it is a defect in the task set I tested against, which declares `environment_mode = "separate"` but ships no `[[verifier.collect]]` step, so the agent's work never reaches the fresh verifier container and it grades unmodified code. With separate-mode grading on, this branch still gets a clean `env_done` and a real verifier run (`reward=0 f2p_passed=0/38`) rather than the previous `grading_error` — the image fix is what moved it from "infrastructure broke" to "verifier ran and scored".

Also worth noting for anyone reading the numbers: that task's grader reported `via=aggregate` with `parsed=0`, i.e. suite-level pass/fail rather than per-test attribution. Fine for an oracle differential, but the 38/38 is derived from the suite exiting clean, not from parsing 38 individual results.

PR #843, opened 2026-08-13
fix(data): preserve think-tag training masks

signalrush:
Part 3 of 13 in the draft stack replacing #809.

Base: `agent/sft-wire-normalization`
Previous: #842
Next: #844

## Change

Honor explicit trainable masks when exploding or retaining think-tag content. Complete masks are preserved; partial masks still use normal derivation.

## Verification

- focused bridge and dataset CLI tests
- Full stack tip: 167 passed, 22 skipped

This PR is intentionally limited to one reviewable behavior.


signalrush:
Folded into #842, which now implements the complete canonical SFT import contract in one targeted commit.

PR #844, opened 2026-08-13
fix(sft): reject ambiguous inline thinking

signalrush:
Part 4 of 13 in the draft stack replacing #809.

Base: `agent/sft-think-mask-preservation`
Previous: #843
Next: #845

## Change

Reject ambiguous structural inline thinking before hosted SFT can silently reinterpret it. Verl explicitly retains its legacy inline-text compatibility.

## Verification

- focused schema, bridge, and dispatcher tests
- Full stack tip: 167 passed, 22 skipped

This PR is intentionally limited to one reviewable behavior.

### DIFF
--- a/rllm/data/sft_bridges.py
+++ b/rllm/data/sft_bridges.py
@@ -45,6 +45,7 @@ from rllm.data.sft_schema import (
     SFTMessage,
     SFTRow,
     SFTSchemaError,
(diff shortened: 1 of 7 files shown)


signalrush:
Folded into #842. The consolidated implementation keeps ambiguity checks at the import boundary and removes backend-specific schema policy.

PR #846, opened 2026-08-13
fix(renderers): normalize Tinker wire history

signalrush:
Part 4 of 11 in the draft stack replacing #809.

Base: `agent/sft-gateway-tool-calls`
Previous: #845
Next: #847

## Change

Normalize structured content, provider reasoning, tool-call objects, and OpenAI tool declarations at the Tinker renderer boundary. Qwen history handling uses the last genuine user message rather than a tool-response surrogate.

## Why it matters

This keeps multi-message reasoning and tool histories semantically intact when adapting between harness/OpenAI wire values and tinker-cookbook renderers.

## Verification

- `pytest tests/test_renderers.py -q`
- full stacked tip: 152 passed, 22 skipped

PR #847, opened 2026-08-13
fix(sft): align training renderer with serving

signalrush:
Part 5 of 11 in the draft stack replacing #809.

Base: `agent/sft-renderer-wire-history`
Previous: #846
Next: #848

## Change

Route row-level tools and the history-reasoning policy through the same renderer behavior used during serving.

## Why it matters

SFT must tokenize the same semantic conversation that inference will render. This prevents silent train/serve drift for structured thinking, tool declarations, tool calls, and multi-message history.

## Verification

- `pytest tests/trainer/test_sft_render_path.py -q`


signalrush:
Superseded by #845, which consolidates the complete rendering and train/serve contract, including production-renderer parity and fail-closed target attribution.

PR #849, opened 2026-08-13
fix(sft): validate on completed-step boundaries

signalrush:
Part 7 of 11 in the draft stack replacing #809.

Base: `agent/sft-forward-only-validation`
Previous: #848
Next: #850

## Change

Run optional initial validation at step 0, then validate and save on completed-update boundaries. Drain outstanding submitted training work before every validation boundary.

## Why it matters

Metrics and checkpoints now describe completed optimizer updates rather than submitted/off-by-one work, and validation cannot race an in-flight update.

## Verification

- focused validation-boundary tests


signalrush:
Superseded by #848, which keeps forward-only validation and completed-step boundary semantics in one reviewable contract.

PR #852, opened 2026-08-13
feat(sft): expose optimizer controls

signalrush:
Part 10 of 11 in the draft stack replacing #809.

Base: `agent/sft-training-horizon`
Previous: #851
Next: #853

## Change

Share validated LR warmup, minimum LR, Adam betas/epsilon, weight decay, and gradient clipping controls across Tinker and Fireworks. Defaults preserve the former effective behavior.

## Why it matters

The hosted backends can express and reproduce the optimizer contract of an external SFT recipe without provider-specific forks.

## Verification

- `pytest tests/trainer/test_sft_lr_warmup.py -q`


signalrush:
Superseded by #850, which consolidates optimizer controls with batching, loss, horizon, and preflight as one hosted training-execution contract.

PR #853, opened 2026-08-13
fix(sft): preflight data before provider launch

signalrush:
Part 11 of 11 in the draft stack replacing #809.

Base: `agent/sft-optimizer-controls`
Previous: #852

## Change

Render the exact planned training batches and every validation batch before creating a Tinker client or provisioning a Fireworks trainer. Preflight restores dataset order after validation.

## Why it matters

Malformed schema, overlength examples, empty supervision, or invalid planning fail before paid provider resources start.

## Verification

- `pytest tests/trainer/test_sft_preflight.py -q`


signalrush:
Superseded by #850, which consolidates provider preflight with batching, loss, horizon, and optimizer semantics as one hosted training-execution contract.

PR #865, opened 2026-08-15
feat(sft): add reproducible 256K Nemotron Super training paths

signalrush:
## Summary

- add a reproducible DeepSWE/Nemotron Super 256K Fireworks SFT launcher and contract
- add a matching Tinker 256K LoRA pilot with token-mean loss, W&B logging, and validation/save cadence controls
- harden hosted SFT checkpoint/resume, validation, training-horizon, and optimizer telemetry semantics

## Validation

- trainer semantics suite: 75 passed
- targeted terminal-RL no-op harness/engine tests passed
- Tinker launcher dry-run passed using the pinned local tokenizer cache

## Notes

The Tinker launcher defaults to a bounded 25-step pilot. A full run requires an explicit opt-in environment variable.

### DIFF
--- a/cookbooks/swe-rl/deepswe_nemotron3_super_256k_fireworks.yaml
+++ b/cookbooks/swe-rl/deepswe_nemotron3_super_256k_fireworks.yaml
@@ -1,0 +1,26 @@
+# Fireworks Training API overrides for Nemotron 3 Super 120B-A12B at 256K.
+# The model, tokenizer, and LoRA training shape must move together.
+
+model:
+  tokenizer_model: nvidia/NVIDIA-Nemotron-3-Super-120B-A12B-BF16
+
+fireworks_config:
+  policy_trainer_shape_id: accounts/fireworks/trainingShapes/nemotron-3-super-120b-a12b-bf16-262k-lora
+  policy_trainer_replica_count: 1
+
+data:
+  rllm:
+    overlength_policy: error
+    loss_reduction: token_mean
+
+optim:
+  betas: [0.9, 0.999]
+  eps: 1.0e-8
+  warmup_steps_ratio: 0.1
+  warmup_steps: -1
+  min_lr: 0.0
+  weight_decay: 0.01
+  grad_clip_norm: 1.0
+
+trainer:
+  max_steps: null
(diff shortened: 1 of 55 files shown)


signalrush:
Rebased flat onto latest `terminal-rl` (a5fa9aa1): the 20-commit history is now **one commit**, conflicts with #867/#870/#880 resolved by adopting the upstream contracts rather than re-fighting them.

**Dropped as superseded:**
- in-`fit()` startup preflight (both backends) + `iter_preflight_batches` and its test — #867 made preflight an explicit `rllm sft --preflight` step and removed the `planned_batches` path this called
- `tests/trainer/test_sft_training_horizon.py` fell out of the diff entirely — after removing the preflight test it is byte-identical to upstream

**Adapted:**
- `test_raw_row_cursor_round_trips_partial_batches_and_epochs` now passes `drop_last=False` (its subject is partial batches, which #870's drop-incomplete default removed); a new sibling test covers the cursor round-trip under the default

**Unchanged — the feature set under review:**
- `SFTResumeContract`: dataset content fingerprint + optimizer/config digest, validated on restart; `data_cursor_for_step`/`step_for_data_cursor` invert at batch boundaries
- `iter_training_batches_from_step` for mid-epoch resume
- Fireworks optimizer telemetry + truncated-model-ID sanitation
- the 256K Nemotron Super cookbook (yaml ×2, prep script, launch scripts ×2)

**Validation:** `tests/trainer` + `tests/cli` (minus optional-verl collection): 300 passed; the 5 failures are byte-identical on clean `terminal-rl` (stash/rerun) and unrelated. Ruff check + format clean.

signalrush:
Stripped further per review: **cookbook only — 5 files, +1,151, zero trainer changes.**

Rationale: upstream already resumes both backends (Tinker `get_last_checkpoint` → state client; Fireworks server-side `ckpt.resume()`), the launch scripts never invoke the resume-contract API (they call stock `rllm sft` with standard flags), and the recipe's reproducibility protection lives in the scripts' own manifest/canary hashes. Even the model-ID sanitation exists upstream (same regex + 63-char cap; the delta was a trailing-hyphen edge case). None of the backend machinery is load-bearing for this recipe.

The exact-resume contract work (dataset fingerprint + optimizer digest validated on restart, cursor round-trip, its 849-line test suite) is preserved on a separate branch, rebased and green, and can become its own PR where it can be reviewed as the independent feature it is.

PR #868, opened 2026-08-15
gateway: compact trace pipeline — store, fetch, and wiring

signalrush:
## What

Chat-completion calls resend the cumulative conversation every turn, so storing full `TraceRecord`s repeats the same messages and prompt token IDs. This PR adds a compact in-memory store and JSON wire format. Compact storage is now the default for rLLM training; set `rllm.gateway.store=memory` or `RLLM_GATEWAY_STORE=memory` to restore the previous behavior. `RLLM_GATEWAY_STORE` overrides the configured backend, and the legacy endpoint and flat record shape remain unchanged.

## Contract

A `TraceDelta` is either a root or a delta against one completed conversational parent:

```text
child.messages         = parent.messages + [parent.response_message] + messages_suffix
child.prompt_token_ids = parent.prompt_token_ids + parent.completion_token_ids + prompt_ids_suffix
```

The parent response and completion IDs are inherited exactly once. The child's own response, completion IDs, logprobs, routing data, finish reason, weight version, and metadata remain explicit. Raw request/response capture is rejected in compact mode because it would retain the full repeated payload.

`TraceGraph.add()` walks a trie of completed message blocks and accepts a parent only when session, lineage, model, messages, and the completed prompt-token prefix match exactly; otherwise the trace becomes a root. `resolve()` reconstructs one flat record, `flatten()` reconstructs a session, and `slice()` rebases a boundary trace. A retry reusing a trace ID replaces only a leaf.

`MemoryTraceStore(compact=True)` owns one graph per session. Compact-capable clients can fetch the graph directly; legacy callers continue to receive flattened records. Ingest still receives cumulative requests, so this PR removes repeated retained state and wire payloads rather than claiming sublinear input scanning.

## Verification

- Final focused stack suite: 180 passed, 1 skipped; default/override selection tests include the real shared YAML and `RLLM_GATEWAY_STORE=memory`.
- Real-data end-to-end replay: 8 sessions, 175 traces, 4,909,670 cumulative prompt IDs; default store, compact store, compact JSON wire, StepDelta, TrajectoryDelta, Tinker Datum, and VERL DataProto views matched exactly. Compact prompt suffixes retained 201,022 IDs.
- Earlier storage benchmark on the same corpus: legacy wire 35.9 MB vs compact 5.1 MB (7.04x smaller); verified-prefix reuse improved compact ingest 255 ms to 158 ms.

## Stack

- #871 maps `TraceDelta` to `StepDelta`.
- #874 adds the minimal `TrajectoryDelta` container.
- #876 feeds compact trajectories into the maintained Tinker and VERL consumers.


signalrush:
**Exhaustive parity sweep complete** (`RLLM_PARITY_LIMIT=0`): every episode of every local eval dump replayed through both store modes —

> **3,979 real episodes byte-identical in both modes** (59 run dirs, ~90 GB: DeepSWE full runs, terminal-bench incl. a 2,000-episode pass-at-16 dump, oracle runs, throttle-mangled and non-cumulative episodes). `1 passed in 3485.95s`.

No non-identical byte found.

signalrush:
**Strengthened parity evidence — old-code A/B protocol.** Beyond the input-referenced parity tests, the store/fetch path is now verified against the *genuinely executed* pre-refactor code: the unchanged base-commit store (run from a pinned worktree) recorded reference outputs for 77 real-episode replays (messages + synthetic prompt-id chains); the compact store + wire fetch + client expansion reproduce those outputs **byte-identically, 77/77**. Fingerprints are now deliberately order-sensitive so equal-content messages with different key order stay distinct nodes — reads reproduce exactly what was stored, never a semantically-equal reordering.

signalrush:
**Review response — all four store/fetch blockers fixed at this head, each with a regression test encoding the reviewer's exact repro:**

1. **Docker hostname**: `host.docker.compactal` → `host.docker.internal` restored (rename collateral); docker/loopback rewrite tests pass again.
2. **Delta cycles**: prompt-id anchors are now keyed per `(session, lineage)`; a replay re-store pins its delta dependents to verbatim ids before overwriting (dependency-safe) and never chains; `_full_prompt_ids` carries a visited-set guard that raises `ValueError` instead of hanging the event loop. The reviewer's t1→t2→t1 repro now returns correct ids for both traces.
3. **`since=` anchors**: compact fetch rebases any trace whose delta ancestor was filtered out to full ids — payloads are always self-contained. Repro returns `[[1,2,3],[1,2,3,4]]` instead of raising.
4. **Client sharing** (partial for the Θ(T²) finding): compact payloads now construct `TraceRecord.model_construct` (gateway-produced, validation-copy skipped), so cross-trace message-dict sharing survives into records — verified end-to-end in #874's integration test. The remaining quadratic (per-trace full id/list materialization) is acknowledged and deferred to the native-compact client/engine representation (review recommendation 2) as a follow-up.

Also: `store=compact` + `db_path` now rejected. 19 store unit tests incl. the four new repros; old-code A/B re-run at this head: 74/74 byte-identical.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #869, opened 2026-08-15
gateway: compact trace fetch — ship each unique message once over the wire

signalrush:
**Stacked on #868** (compact trace store) — second of the store → fetch → export series.

## What

`GET /sessions/{id}/traces?format=compact` returns `{format, nodes, traces}` — a node table plus per-trace `messages_ref` leaf refs — instead of the legacy list whose messages repeat every conversation prefix (quadratic in turns; up to **~400 MB JSON for one 229-step task**). The **default format is untouched and remains the default**.

## Design

- **Format independent of store mode**: a compact-mode store serves from its node table; a default-mode store interns ephemerally at read time. Stores without compact support (sqlite) serve the legacy form regardless of the requested format — safe against older gateways.
- **Clients expand transparently**: both `GatewayClient` and `AsyncGatewayClient` accept `format="compact"`, rebuild full message lists before constructing `TraceRecord`s — callers see the exact same return shape. Expansion materializes each node **once and shares it across traces**, so client memory stays linear in unique messages.

## Parity — real data, not mocks

`test_real_dump_fetch_parity_compact_vs_default` replays real eval episodes through store → compact payload → **JSON wire round-trip** → client expansion:

- **152 sampled episodes across all local dumps** (~44 run dirs: DeepSWE + terminal-bench, throttle-mangled and non-cumulative included): expansion is canonical-byte-identical to both the original conversations **and** the default fetch, in **both** store modes — pass
- unit test additionally pins prefix sharing (`expanded[1][0] is expanded[0][0]`)

### DIFF
--- a/rllm-model-gateway/src/rllm_model_gateway/client.py
+++ b/rllm-model-gateway/src/rllm_model_gateway/client.py
@@ -36,6 +36,7 @@ def _expand_compact_traces(payload: dict[str, Any]) -> list[dict[str, Any]]:
             paths[nid] = base
         return paths[leaf] if leaf is not None else []
 
+    ids_memo: dict[str | None, list[int]] = {None: []}
     for trace in payload.get("traces", []):
         ref = trace.get("messages_ref")
         if ref is None:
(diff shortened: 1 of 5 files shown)


signalrush:
Consolidated into #868, which now carries the whole gateway compact pipeline (store + fetch + wiring) as one cohesive PR. Content unchanged — the final tree of the consolidated stack is byte-identical to this stack's tip.

PR #871, opened 2026-08-15
types: add completed-parent StepDelta

signalrush:
Stacked on #868.

## What

`StepDelta` carries the compact gateway contract into the training type layer without re-storing cumulative conversations:

```text
child.prompt_ids       = parent.prompt_ids + parent.response_ids + child.prompt_ids_suffix
child.chat_completions = parent.chat_completions + child.chat_completions_suffix
```

Roots store their inputs in full. Children name their completed parent with `parent_step_id`, so interleaved lineages and sibling branches resolve by explicit ID rather than list position. The direct `TraceDelta -> StepDelta` mapper retains the small fields needed to reconstruct the legacy gateway `ModelOutput`, including `finish_reason`, without duplicating the growing prompt. It also preserves either `reasoning` or OpenAI-compatible `reasoning_content` exactly.

#876 wires this contract into the live training path; this PR defines and tests the type, mapper, and resolver.

## Verification

- Final focused stack suite: 180 passed, 1 skipped; ruff clean.
- Real token-heavy replay: 175/175 gateway-produced Steps exact after `TraceGraph -> StepDelta -> JSON -> Step`, covering 4,909,670 cumulative prompt IDs and 201,022 stored suffix IDs.
- The exactness contract is scoped to canonical gateway-produced Steps; arbitrary non-gateway `ModelOutput` extensions are outside the compact representation.


signalrush:
**Strengthened parity evidence — file-byte protocol.** The parity test now enforces a stricter property than canonical-JSON equality: a legacy episode file converted to compact and back must reproduce the **exact bytes** the previous pipeline wrote (same writer settings). This immediately caught a real defect the canonical test had certified as lossless — `expand` re-appended `chat_completions` at the end of each step dict, and canonical (sorted-key) comparison hid the order change. Fixed by swapping marker *values* in place (key order survives the round trip) and making fingerprints order-sensitive. Result on real dumps, including the mixed-key-order terminal-bench dirs that exposed it: **443/443 files byte-identical**.

signalrush:
**Review response — episode-layer findings fixed at this head:**

- **Cycle/dangling guards**: the expand walk keeps a visited set and raises `ValueError` ("cycle" / "dangling") instead of hanging curation or a viewer request; tests forge both corruptions.
- **Reserved-key losslessness**: compaction refuses (stays legacy, still lossless) when `episode_format` exists with another value or a trajectory already carries `message_nodes`; expand passes refused trajectories through untouched. A zero-step trajectory no longer gains an empty node table (caught by the byte-parity harness immediately after the refusal fix — the strict harness continues to earn its keep).
- **Hermetic test**: the real-dump parity scan is now opt-in (`RLLM_PARITY_SCAN=1` or `RLLM_PARITY_DUMPS`) with a byte budget (`RLLM_PARITY_MAX_MB`, default 4 GB); default runs are synthetic-only.

Open per review recommendation 6 (follow-up): serving compact episodes to the browser with per-step lazy materialization — the server-side expansion for legacy rendering remains the current behavior.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #885, opened 2026-08-17
harbor: make served model names litellm-routable for sandboxed agents

signalrush:
## The bug

Harbor agents (OpenHands SDK et al.) call the model through **litellm inside the sandbox**, which parses the text before the first `/` as the provider. rLLM's `build_harbor_trial_config` only prefixes slash-less model names; slash-bearing names pass through verbatim. So a served name like `nvidia/nemotron-3-ultra-550b-a55b` dies before any HTTP request:

```
litellm.BadRequestError: LLM Provider NOT provided. Pass in the LLM provider you are trying to call.
You passed model=nvidia/nemotron-3-ultra-550b-a55b
```

`nvidia` is not a litellm provider (their NVIDIA adapter is `nvidia_nim`), so the agent never reaches the inference tunnel — observed live on `rllm eval harbor:swebench-verified --agent <openhands-sdk> --model nvidia/nemotron-3-ultra-550b-a55b --sandbox-backend modal`.

## The fix

Wrap names whose first segment is **not** in `litellm.provider_list` as `openai/<name>`. litellm then selects its OpenAI adapter, routes to `LLM_BASE_URL` (rLLM's proxy, via the ngrok tunnel on remote sandboxes), and sends the **full** name on the wire — exactly the name the proxy serves, so the agent-side request and the proxy's `model_list` line up with the user's single `--model` value unchanged.

Untouched by design:
- names with real provider prefixes (`anthropic/…`, `openai/…`, `openrouter/…`, `hosted_vllm/…`) keep their existing routing — `anthropic/` continues through the proxy's `/v1/messages` passthrough, `hosted_vllm/` keeps its dummy-key handling
- bare names keep the existing `_infer_provider_prefix` inference
- the training placeholder (`openai/placeholder`) — first segment is a known provider

5 unit tests: the failing name wraps, known prefixes untouched, wrapping idempotent, bare-name inference unchanged, placeholder unchanged.

### DIFF
--- a/rllm/integrations/harbor/trial_helper.py
+++ b/rllm/integrations/harbor/trial_helper.py
@@ -104,6 +104,29 @@ def _infer_provider_prefix(model_name: str) -> str:
     return f"openai/{model_name}"
 
 
+def _litellm_routable(model_name: str) -> str:
+    """Make *model_name* parseable by the litellm inside the Harbor agent.
+
+    litellm reads the text before the first "/" as the provider. Served model
+    names like ``nvidia/nemotron-3-ultra-550b-a55b`` carry a first segment that
+    is NOT a litellm provider, so the agent fails before making any request
+    ("LLM Provider NOT provided"). Wrapping as ``openai/<name>`` routes the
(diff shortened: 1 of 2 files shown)


signalrush:

PR #891, opened 2026-08-19
debug: dump raw TraceRecords and compact graphs for parity checks

signalrush:
## Why

This is debug-only instrumentation for validating the compact gateway trace pipeline against real training traffic, including the workload from [W&B run z6xi3xkh](https://wandb.ai/agentica/terminal-rl/runs/z6xi3xkh).

The two artifacts are intentionally independent:

- `raw_trace_records.jsonl` is serialized directly from the full `TraceRecord` inside `MemoryTraceStore.store_trace()`, before `TraceGraph.add()` or `replace_leaf()` runs.
- `trace_graph.json` is the eventual compact graph emitted later on full retrieval or session deletion.

The raw file is **not** produced from `TraceGraph.flatten()`, an `Episode`, a `Step`, or a `Trajectory`. Deriving both sides from the graph would make parity tautological and hide conversion bugs.

## What changed

- Add opt-in `rllm.gateway.trace_parity_dump_dir` / `RLLM_GATEWAY_TRACE_PARITY_DUMP_DIR` plumbing for thread and multi-worker gateway modes.
- Dump full pre-compaction records and the eventual graph into per-session, per-generation directories.
- Keep retry incarnations of the same session separate and preserve every raw replacement input.
- Add `scripts/verify_trace_parity_dump.py`.
  - Independently derive the expected completed-message parent from raw records.
  - Check `parent_trace_id`, `messages_suffix`, and `prompt_ids_suffix`.
  - Then check exact full-record reconstruction.
  - Handle same-leaf replacement semantics for retried trace IDs.
- Add a regression test proving that a lossless all-root graph passes `flatten()` equality but is still rejected as structurally wrong.

## z6xi3xkh reproduction

The W&B run points to commit `d501e357967c45bb46807d5f6a3366cbcdfee32b` on `kyle/terminal-rl-fireworks`. I checked out that exact commit and compared it with the current `terminal-rl` base. The current cookbook uses Terminus-2, so this PR restores the run's `MiniSweAgentHarness` entrypoint and the run-required MiniSwe exit-status/config behavior instead of silently changing its request stream. The captured W&B environment used `transformers==5.15.0`, which the launcher pins explicitly because this branch's installed 5.5.3 does not recognize `model_type=deepseek_v4`.

`cookbooks/terminal-rl/reproduce_z6xi3xkh.sh` carries over the W&B-recorded model, LoRA rank, Fireworks shapes/replicas, group and async sizes, optimizer, rollout limits, DPPO-TV settings, and MiniSwe limits. Intentional current-machine/current-branch differences are documented at the top of the script:

- current dataset registry name `tb-v2-debug`
- compact store plus independent raw-record/graph dumps
- automatic tunnel resolution instead of the old machine's fixed gateway route
- shorter experiment name to avoid the original run's Fireworks output-model-id length failure

The current machine has an ngrok wildcard registered. Leave `rllm.gateway.tunnel` and `rllm.gateway.port` unset: `AgentTrainer` creates a unique wildcard hostname and free gateway port for the run. `RLLM_GATEWAY_TUNNEL` remains an optional explicit override.

Review the launch without provisioning or training:

```bash
RLLM_LAUNCH_DRY_RUN=1 bash cookbooks/terminal-rl/reproduce_z6xi3xkh.sh
```

Parse and print the resolved static Hydra config:

```bash
RLLM_LAUNCH_CONFIG_ONLY=1 bash cookbooks/terminal-rl/reproduce_z6xi3xkh.sh
```

The launcher defaults the dump directory to `/data/home/thw/trace-dumps/z6xi3xkh-parity-001`, so it can be run directly. `RLLM_GATEWAY_TRACE_PARITY_DUMP_DIR` remains available as an override for a later run. For a bounded first real launch, append `rllm.trainer.total_batches=1`. The launcher automatically runs the parity verifier after training exits, including failed runs.

## Operational warning

The dump feature is off by default outside this reproduction launcher. It is compact-store-only, synchronous, disk-heavy, and may contain sensitive prompts/token data. It is intended for small debugging runs on fast local storage, not normal training or performance measurements.

A first manual launch exposed the Transformers version mismatch after Fireworks provisioning but before any rollout or trace generation; Fireworks deleted that trainer job cleanly. The pinned 5.15.0 environment now loads the exact DeepSeek V4 config and tokenizer locally before another launch.

## Validation

- `rllm-model-gateway/.venv/bin/python -m pytest rllm-model-gateway/tests/unit -q` — 308 passed
- `PYTHONPATH=. uv run pytest -q tests/harnesses/test_cli_harness.py tests/engine/test_agentflow_engine.py tests/engine/test_gateway_manager.py tests/engine/test_trace_parity_dump.py` — 122 passed
- launcher dry run — passed without requiring a per-run tunnel
- Hydra config-only parse — passed
- exact `deepseek-ai/DeepSeek-V4-Flash-0731` config/tokenizer load under pinned Transformers 5.15.0 — passed (`DeepseekV4Config`, max positions 1048576, vocabulary 128000)
- focused synthetic raw-record -> graph -> verifier smoke test — passed
- repository pre-commit hooks (`ruff`, `ruff-format`) — passed
- `git diff --check` — passed

### DIFF
--- a/cookbooks/terminal-rl/reproduce_z6xi3xkh.sh
+++ b/cookbooks/terminal-rl/reproduce_z6xi3xkh.sh
@@ -5,6 +5,8 @@
 # Source of truth:
 #   - W&B recorded program/config from commit d501e357967c45bb46807d5f6a3366cbcdfee32b
 #   - W&B run arguments are copied below field-for-field
+#   - W&B captured transformers==5.15.0; pin it because this branch's 5.5.3
+#     does not recognize model_type=deepseek_v4
 #
 # Intentional current-branch adaptations:
 #   - tb_v2_debug -> the current registry name tb-v2-debug
@@ -74,7 +76,7 @@ export RLLM_SANDBOX_TIMEOUT_S="${RLLM_SANDBOX_TIMEOUT_S:-3000}"
 export RLLM_EXPERIMENT_NAME="${RLLM_EXPERIMENT_NAME:-z6xi3xkh-dsv4-flash-lora128-trace-parity}"
 
 launch_cmd=(
-    uv run python -u cookbooks/terminal-rl/train_debug.py
+    uv run --with transformers==5.15.0 python -u cookbooks/terminal-rl/train_debug.py
     rllm/backend=fireworks
     model.name=accounts/fireworks/models/deepseek-v4-flash-0731
     model.tokenizer_model=deepseek-ai/DeepSeek-V4-Flash-0731
--- a/cookbooks/terminal-rl/reproduce_z6xi3xkh.sh
+++ b/cookbooks/terminal-rl/reproduce_z6xi3xkh.sh
@@ -1,0 +1,172 @@
+#!/usr/bin/env bash
+# Reproduce https://wandb.ai/agentica/terminal-rl/runs/z6xi3xkh on the
+# current terminal-rl branch, with raw TraceRecord -> TraceGraph parity dumps.
+#
+# Source of truth:
+#   - W&B recorded program/config from commit d501e357967c45bb46807d5f6a3366cbcdfee32b
+#   - W&B run arguments are copied below field-for-field
+#
+# Intentional current-branch adaptations:
+#   - tb_v2_debug -> the current registry name tb-v2-debug
+#   - the old machine's fixed http://5.78.144.17:19090 route is replaced by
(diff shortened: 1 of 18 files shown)


listar2000:
@signalrush Might be a good chance to try the new `skills/rllm-simplify/SKILL.md` for these 3 recent PRs (including #892 #893 )

PR #898, opened 2026-08-22
fix(engine): drop logprobs when OpenAIEngine retokenizes the completion text

### LINKED ISSUE #816: OpenAIEngine.completion() retokenization fallback can desync completion_ids from logprobs
### Description

`OpenAIEngine.completion()` (`rllm/engine/rollout/openai_engine.py`, lines ~184-202) falls back to `completion_ids = self.tokenizer.encode(text, add_special_tokens=False)` whenever the response has no usable `token_ids` (the `assert completion_ids is not None` a few lines above fails, e.g. because the backend doesn't return the non-standard vLLM `token_ids` field). `logprobs` is still taken from the server's original `response.choices[0].logprobs.token_logprobs`, which is aligned to the tokens the model actually sampled, not to the retokenized text.

Autoregressive sampling does not guarantee `encode(decode(ids)) == ids` for a BPE tokenizer, so this fallback can produce a `completion_ids` list whose length or per-index alignment no longer matches `logprobs`, silently.

Downstream, `rllm/trainer/verl/transform.py` (`_new_segment` and the segment-extension loop, ~lines 316-324 and 378-381) already contains a workaround for exactly this: when `len(action_lp) != len(action)`, it pads `action_lp` with `0.0` at the tail, with the comment "If logprobs missing/short, pad to action length with zeros so accumulator lists stay aligned." That avoids a shape crash but doesn't restore correct alignment, so training would use wrong (zeroed) old-logprobs for some tokens instead of failing loudly.

### Steps to Reproduce

This is a source-level finding rather than a crash, so there's no training script that fails outright. What I reproduced is the underlying tokenizer mechanism the fallback silently depends on:

1. In a clean `python:3.12-slim` container: `pip install tiktoken`
2. ```python
   import tiktoken
   enc = tiktoken.get_encoding("cl100k_base")
   ids = [0, 0]
   text = enc.decode(ids)                              # '!!'
   re_ids = enc.encode(text, disallowed_special=())
   print(re_ids)                                        # [3001], not [0, 0]
   ```
3. `encode(decode(ids)) != ids` on the very first pair of vocab ids tried, confirming decode-then-re-encode is not an identity operation for a BPE tokenizer in general.

I didn't have network/model access to run rLLM's own HF tokenizer against a live server that omits `token_ids`, so I can't show the exact desync happening inside `OpenAIEngine.completion()` end to end. The mechanism is demonstrated generically here and corroborated by the pad-to-align workaround already in `transform.py`, which only makes sense as a fix if this happens in practice.

### Error Output / Traceback

None. It doesn't crash; it produces silently misaligned `logprobs`, currently only masked downstream by the zero-pad in `transform.py`.

### rLLM Version

commit `1d1109a655e291b3001d8526d7c9ecc5b9328226` (current main)

### Training Backend

N/A (`OpenAIEngine.completion()` itself; both verl and tinker paths that use this engine route through it)

### Python Version

3.12

### Additional Context

This might be related to #331 (abnormal outputs in later turns of multi-turn verl training), since wrong old-logprobs would corrupt PPO/GRPO importance-sampling ratios in a way that could look like training instability, but I haven't verified that connection and I'm not confident it's the same root cause.

Happy to send a PR if useful, for example clearing/dropping `logprobs` when the `token_ids` fallback fires (since they no longer correspond to the retokenized ids), but I wanted to flag the tradeoff first since there might be a better fix you'd prefer (e.g. requesting `token_ids` more strictly, or masking just the drifted tail instead of zeroing).

AmirF194:
## Summary

`OpenAIEngine.completion()` retokenizes the decoded completion text whenever the
server does not return the vLLM-only `token_ids` field, but keeps pairing the
server's original per-token `logprobs` with those retokenized ids anyway. The two
are no longer guaranteed to correspond: `encode(decode(ids)) == ids` does not hold
in general for BPE-style tokenizers, so `completion_ids` and `logprobs` can end up
different lengths or misaligned per-index.

The existing downstream workaround in `rllm/trainer/verl/transform.py` zero-pads a
short `logprobs` list to the action length, which stops the shape crash but silently
feeds a wrong (zero) old-logprob into PPO/GRPO importance-sampling ratios for the
tail tokens instead of surfacing the mismatch.

## Root cause

`rllm/engine/rollout/openai_engine.py`, `completion()`: when
`response.choices[0].token_ids` is absent, `completion_ids` comes from
`self.tokenizer.encode(text, ...)` (a fresh retokenization), while `logprobs` still
comes from `response.choices[0].logprobs.token_logprobs` (the server's per-token
values for whatever it actually sampled). Nothing ties the two together once the
fallback fires.

## Fix

Track whether `completion_ids` came from the native `token_ids` field or the
retokenization fallback. When it's the fallback, clear `logprobs` to `[]`, matching
the "logprobs unavailable" signal the code already produces when
`response.choices[0].logprobs` itself is missing. This keeps the single place that
already interprets an empty `logprobs` list correctly, rather than adding a second
code path.

## Validation

- New regression test `tests/engine/test_openai_engine_completion.py`: fails on
  `main` (`assert [-0.1, -0.2] == []`) and passes on this branch, for both the
  fallback case (logprobs cleared) and the native-`token_ids` case (logprobs
  preserved unchanged). Ran with `pytest`, no network, in a clean
  `python:3.12-slim` container, module `__file__` printed and confirmed pointing
  at the checked-out source.
- `ruff check` and `ruff format --check` (pinned `0.11.4`, matching
  `.pre-commit-config.yaml`) clean on both changed files.
- Not run: this repo's CI only exercises `openai_engine.py` changes through
  `test-tinker.yml`, which is path-scoped to run `tests/engine/test_tinker_engine.py`
  only (verified by reading the workflow), so the new test above is not wired into
  any existing CI job; I did not add a workflow entry for it since that's a repo
  process choice. No live OpenAI-compatible endpoint was queried; the response
  shapes in the test are constructed by hand from the fields `completion()` reads.

Fixes #816

### DIFF
--- a/rllm/engine/rollout/openai_engine.py
+++ b/rllm/engine/rollout/openai_engine.py
@@ -186,8 +186,10 @@ class OpenAIEngine(RolloutEngine):
                 try:
                     completion_ids = response.choices[0].token_ids
                     assert completion_ids is not None
+                    completion_ids_are_native = True
                 except Exception:
                     completion_ids = self.tokenizer.encode(text, add_special_tokens=False)
+                    completion_ids_are_native = False
 
                 parsed_output = self.chat_parser.parse_completion(completion_ids)
 
@@ -199,6 +201,13 @@ class OpenAIEngine(RolloutEngine):
                     assert response.choices[0].logprobs is not None
                     logprobs = response.choices[0].logprobs.token_logprobs
                 except Exception:
+                    logprobs = []
+
+                if not completion_ids_are_native:
+                    # The server's logprobs are aligned to the tokens it actually sampled,
+                    # not to completion_ids retokenized from the decoded text above
+                    # (encode(decode(ids)) == ids is not guaranteed), so pairing them here
+                    # would silently desync per-token training signal downstream.
                     logprobs = []
 
                 if sampling_params.get("echo", False) and logprobs:
--- a/tests/engine/test_openai_engine_completion.py
+++ b/tests/engine/test_openai_engine_completion.py
@@ -1,0 +1,78 @@
+"""Regression tests for OpenAIEngine.completion()'s token_ids retokenization fallback."""
+
+import asyncio
+from types import SimpleNamespace
+from unittest.mock import AsyncMock
+
+from rllm.engine.rollout.openai_engine import OpenAIEngine
+
+
+class _FakeTokenizer:
+    """Always returns a fixed, deliberately non-native-length token stream.
+
+    Stands in for a real tokenizer where encode(decode(ids)) != ids, which is common
+    (not a rare edge case) with BPE-style tokenizers.
+    """
+
+    def encode(self, text, add_special_tokens=False):  # noqa: ARG002
+        return [101, 102, 103]
+
+
+class _FakeChatParser:
+    def parse_completion(self, completion_ids):  # noqa: ARG002
+        return {"content": "", "reasoning": None, "tool_calls": []}
+
+
+def _make_engine():
+    engine = OpenAIEngine(
+        model="test-model",
+        tokenizer=_FakeTokenizer(),
+        chat_parser=_FakeChatParser(),
+        max_prompt_length=64,
+        max_response_length=64,
+        base_url="http://localhost:0/v1",
+        api_key="test",
+    )
+    engine.sampling_params = {}
+    return engine
+
+
+def _completions_response(*, token_ids, logprobs):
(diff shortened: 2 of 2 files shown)


AmirF194:
Checking in a week after opening this one, no rush. Let me know if you'd like anything reworked.

AmirF194:
Closing to keep the queue clean; happy to reopen if there's interest.

PR #909, opened 2026-09-05
fix: preserve hierarchical task IDs in GRPO grouping

### LINKED ISSUE #829: Dataset row ids containing ':' silently collapse every rollout into one GRPO group
## Summary

An integration whose dataset row ids contain `:` will silently get **one GRPO group for the entire batch**, because `Episode.task_id` is parsed as everything before the first colon. Nothing errors, every metric stays populated, and advantages are computed across unrelated tasks.

We hit this on a real training run and lost a day to it, so filing in case it is worth a guard.

Observed against `1d1109a655e291b3001d8526d7c9ecc5b9328226`.

## The mechanism

`rllm/types.py:333`

```python
@property
def task_id(self) -> str:
    return self.id.split(":")[0]
```

`rllm/trainer/algorithms/transform.py:127`

```python
trajectories_by_name[f"{task_id}:{trajectory.name}"].append(trajectory)
```

So `:` is a reserved field separator in `Episode.id`, and the grouping key is derived from the segment before the first one. That is a sensible convention, but it is implicit: nothing validates it, and an id that happens to use `:` as a readable hierarchy separator collapses every row to the same `task_id`.

Our row ids looked like:

```
aeread:integration-v1:case01:s1200
```

which yields `task_id == "aeread"` for every row in the dataset.

## What it looks like when it happens

16 distinct dataset rows, 4 case families at 4 seeds each, `rollout.n=2`:

```
groups/num_trajs_before_filter:  32
groups/num_groups:                1
groups/avg_group_size:         32.0
batch/num_tasks:                  1
```

`groups/num_groups: 1` and `batch/num_tasks: 1` are the tells, and they are the only ones. The run completes, `grad_norm` is non-zero, `informative` is non-zero, and the numbers look healthy. In our case a single rollout that scored well was advantaged against 31 rollouts drawn from **different tasks with different reward scales**, which is exactly the comparison GRPO grouping exists to prevent.

Our reward is a normalized ratio whose per-task scale is only safe *because* grouping is per task, so the pooling quietly invalidated the reward design while every metric kept reporting.

Fix on our side was one line, changing the delimiter to `.`:

```
aeread.integration-v1.case01.s1200
```

after which `num_groups` went 1 -> 16 and `avg_group_size` 32 -> 2, with no other change.

## Suggestions, in rough order of cost

1. **Document the reserved delimiter** where dataset rows are described. A sentence saying `id` must be `<task_id>:<rollout_idx>` and that `task_id` must not contain `:` would have prevented this entirely.
2. **Warn when a batch collapses.** In `_build_trajectory_groups`, if the number of distinct `task_id`s is 1 while the episode count is much larger, that is almost always this bug rather than an intended configuration. A one-line log would surface it immediately.
3. **Validate at registration.** `DatasetRegistry.register_dataset` could reject or warn on row ids containing `:`, which catches it before any GPU time is spent.

Happy to send a PR for any of these if useful; (2) seems like the best value for the least intrusion.

## Secondary, lower confidence: reconciliation by position

Related and softer, so treat this as an observation rather than a bug report.

The adapter constructs `Step`s and the gateway records requests separately, then the two are reconciled. That means an extra gateway record relative to recorded turns desynchronizes the correspondence. We had to construct our candidate's OpenAI client with `max_retries=0` specifically so a transparent SDK retry could not produce more gateway records than turns.

That is a fine workaround, but it is a sharp edge: any integration whose client retries internally, or whose harness makes an out-of-band model call, will drift without an obvious error. Keying the correspondence on a response id rather than position would remove the class. We already stamp `Step.id` from the response id on our side, so some of the plumbing exists.

Not asking for anything here, just flagging it since it is adjacent and cost us the same debugging session.

## Context

This came out of building a training integration for an economic-reasoning environment (multi-agent exchange arena, one trainable seat against a frozen panel). Everything above is reproducible from a stock install at the pinned revision; happy to provide a minimal repro if that would help.

houzl3416:
## Summary
Fixes #829: dataset/task IDs containing `:` were left-split in `Episode.task_id` / `TrajectoryGroup` accessors, so GRPO silently merged unrelated tasks into one group.

**Option A (minimal):**
- Parse episode and group IDs with `rsplit(":", 1)` (final colon is the separator)
- Reject default-path trajectory names that contain `:` (avoids ambiguous `a:b:c` serializations)
- Document the contract in `docs/core-concepts/episodes-trajectories-steps.mdx`
- Add CPU regressions in `tests/unified_trainer/test_trajectory_grouping.py` (grouping + real collector GRPO advantages `[-1, 1, -1, 1]`)

Producer wire format (`task_id:rollout_idx`) is unchanged.

## Test plan
- [x] Focused CPU suite reported **33 passed** locally (includes new trajectory-grouping / GRPO regressions)
- [ ] CI pre-commit / review on this PR

### DIFF
--- a/docs/core-concepts/episodes-trajectories-steps.mdx
+++ b/docs/core-concepts/episodes-trajectories-steps.mdx
@@ -66,6 +66,8 @@ class Trajectory(BaseModel):
 
 The `name` field identifies which agent produced this trajectory. During training, trajectories are grouped by `{task_id}:{trajectory.name}` for advantage computation — GRPO compares trajectories within the same group to determine which rollouts were better than average.
 
+When you use default trajectory grouping, trajectory names must not contain `:`. The final colon separates the task ID from the trajectory name.
+
 ## Episode
 
 An Episode is the top-level container for a complete rollout on a single task. Its ID encodes both the task and rollout index.
@@ -82,12 +84,16 @@ class Episode(BaseModel):
     metadata: dict                        # Arbitrary extra data
 ```
 
-The `task_id` and `rollout_idx` properties parse the ID string:
+The `task_id` and `rollout_idx` properties split the ID at the final colon. Task IDs may contain `:`; the final segment is the rollout index:
 
 ```python
 episode = Episode(id="gsm8k_42:3")
 episode.task_id     # "gsm8k_42"
 episode.rollout_idx # "3"
+
+episode = Episode(id="aeread:integration-v1:case01:s1200:0")
+episode.task_id     # "aeread:integration-v1:case01:s1200"
+episode.rollout_idx # "0"
 ```
 
 ## How they connect
--- a/rllm/trainer/algorithms/transform.py
+++ b/rllm/trainer/algorithms/transform.py
@@ -112,6 +112,12 @@ def _build_trajectory_groups(episodes: list[Episode], compact_filtering_config: 
     Returns:
         List of TrajectoryGroups
     """
+    # Group IDs use the final colon to separate the task ID from the trajectory name.
+    for episode in episodes:
+        for trajectory in episode.trajectories:
+            if ":" in trajectory.name:
+                raise ValueError(f"Trajectory name {trajectory.name!r} must not contain ':' when using default trajectory grouping (episode {episode.id!r}).")
+
     trajectories_by_name: dict[str, list[Trajectory]] = defaultdict(list)
     metadata_by_name: dict[str, list[dict]] = defaultdict(list)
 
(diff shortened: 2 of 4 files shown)


houzl3416:
Friendly ping @jeffreysijuntan — CI is green on this one (`pre-commit` / `test`). It fixes #829 (colon in task IDs silently collapsing GRPO groups via left-`split`). Happy to adjust if anything looks off.

jeffreysijuntan:
Great, thanks for taking on this issue.

PR #918, opened 2026-09-10
fix: reap the sandbox process after a code-reward timeout kill

### LINKED ISSUE #915: check_correctness / lcb_check_correctness_v2 leave a zombie process after killing a timed-out test run
## Description

`check_correctness` (`rllm/rewards/code_reward.py:120`) and `lcb_check_correctness_v2` (line 231) run untrusted generated code in a `multiprocessing.Process`, and on timeout do `if process.is_alive(): process.kill()` with no `process.join()` after it. `kill()` only delivers SIGKILL; the child isn't reaped until something calls `join()`/`wait()` on it, so it sits as a zombie (`STAT=Z`) for the life of the worker process.

`lcb_check_correctness_v2` runs on every code-task reward through `RewardCodeFn.__call__` (the LiveCodeBench path and the default fallback), so any slow or hanging generation during RL training leaks one zombie per timeout on that worker. Each one is harmless alone, but a long run with many workers and timeouts can eventually hit the OS per-user PID limit.

## Steps to Reproduce

Call `check_correctness` with a `test_fn` that sleeps past `timeout_per_test`, then check for the spawned child:

```python
import os, subprocess, time
from rllm.rewards.code_reward import check_correctness

def test_fn(*a, **k):
    time.sleep(30)

check_correctness({"inputs": ["1"], "outputs": ["1"]}, "x = 1", test_fn, timeout_per_test=2, max_tests=15)
time.sleep(1)
print(subprocess.run(["ps", "-o", "pid,stat,cmd", "--ppid", str(os.getpid())],
                      capture_output=True, text=True).stdout)
```

## Error Output / Traceback

No exception or crash, that's part of what makes it easy to miss. Ran the snippet above in a clean `python:3.11-slim` container with this repo installed as-is (real `check_correctness`, not a reimplementation):

```
check_correctness returned (False, {...}) after 7.1s
  PID TTY  STAT TIME COMMAND
    1 ?    Ss   0:00 python3 repro.py
  123 ?    Z    0:00 [python3]
```

PID 123 is the killed test-runner subprocess, stuck at `Z`.

## rLLM Version

commit `3b40c37c` (current `main`)

## Training Backend

N/A, reproduces outside the trainer

## Python Version

3.11.5

## Additional Context

`lcb_check_correctness_v2` has the identical pattern at lines 231-234 (`p.kill()`, no `p.join()`); same cause, same fix. The fix looks like adding `process.join()` right after the existing `kill()` call in both functions. Happy to send a PR if useful.

AmirF194:
## Summary

`check_correctness` and `lcb_check_correctness_v2` both `kill()` the sandbox process on
timeout but never `join()` it afterward, so it sits as a zombie until the worker exits.
Added the missing `join()` right after each `kill()`.

## Type of change

- [x] Fix

## What changed

- `process.join()` after `process.kill()` in `check_correctness`
- `p.join()` after `p.kill()` in `lcb_check_correctness_v2`, same pattern

## Validation

- [x] `pre-commit run --files rllm/rewards/code_reward.py tests/rewards/test_code_reward.py`
- [x] Targeted tests: `pytest tests/rewards/test_code_reward.py`

Validation details:
- Two new tests force a timeout with a hanging `test_fn`/`run_test` and read `/proc` for
  the child's state afterward. On main the child is still there at `Z`; with the join()
  added it's gone. The rest of `tests/rewards/test_code_reward.py` is unchanged and green,
  except `test_reward_kodcode`, which needs numpy/pandas in the sandboxed subprocess and my
  container doesn't have those installed, unrelated to this change.

- None

## Docs / examples

- [x] Not needed

## Related issues / PRs

- Fixes #915

One thing from the issue thread I deliberately left out: `how2how2how2-arch` pointed out that
`kill()` only signals the direct child, so if the generated code under test forks or shells
out, those grandchildren get reparented and keep running past this fix. True, and worth doing,
but it's a different change (starting the process in its own session, killing the group), so
I kept this PR to the zombie itself.

### DIFF
--- a/rllm/rewards/code_reward.py
+++ b/rllm/rewards/code_reward.py
@@ -121,6 +121,7 @@ def check_correctness(tests: list[dict[str, str]] | dict[str, list[str]], code: 
 
         if process.is_alive():
             process.kill()
+            process.join()
         test_results_list = list(test_results)
 
         detailed_results: dict[str, Any] = {"all_passed": False, "test_results": [], "total_tests": num_tests, "passed_tests": 0}
@@ -232,6 +233,7 @@ def lcb_check_correctness_v2(sample, generation, timeout=6, debug=False):
 
         if p.is_alive():
             p.kill()
+            p.join()
         if not result:
             in_outs = json.loads(sample["input_output"])
             # consider that all tests failed
--- a/tests/rewards/test_code_reward.py
+++ b/tests/rewards/test_code_reward.py
@@ -1,7 +1,28 @@
+import os
+import time
+
 import pytest
 
+import rllm.rewards.code_reward as code_reward_module
 from rllm.rewards import RewardConfig, RewardType
-from rllm.rewards.code_reward import RewardCodeFn
+from rllm.rewards.code_reward import RewardCodeFn, check_correctness, lcb_check_correctness_v2
+
+
+def _child_process_states(parent_pid):
+    """States of the immediate child processes of parent_pid, read from /proc."""
+    states = []
+    for pid_str in os.listdir("/proc"):
+        if not pid_str.isdigit():
+            continue
+        try:
+            with open(f"/proc/{pid_str}/stat") as f:
+                after_comm = f.read().rsplit(")", 1)[1].split()
+        except (FileNotFoundError, ProcessLookupError):
+            continue
+        state, ppid = after_comm[0], after_comm[1]
+        if ppid == str(parent_pid):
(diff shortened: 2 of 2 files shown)


AmirF194:
This closes a real zombie-process leak in the code-reward sandbox (the killed child stays a zombie with no join(), confirmed live). CI's clean and it's been quiet for a week, flagging in case it slipped past review.

